PR #5, opened 2024-03-22
Introduce `switch` promotion test case and reorder loop promotion passes

andrealinux1:
- Introduce a new FileCheck tests for the dual case `switch` to `if` promotion.
- Postpone the beautify loop promotion passes after the `switch` to `if` promotion, in order to catch more promotion opportunities.

### DIFF
--- a/lib/RestructureCFG/BeautifyGHAST.cpp
+++ b/lib/RestructureCFG/BeautifyGHAST.cpp
@@ -1015,16 +1015,6 @@ void beautifyAST(const model::Binary &Model, Function &F, ASTTree &CombedAST) {
   RootNode = matchSwitch(CombedAST, RootNode);
   Dumper.log("after-switch-match");
 
-  // Match dowhile.
-  revng_log(BeautifyLogger, "Matching do-while\n");
-  matchDoWhile(RootNode, CombedAST);
-  Dumper.log("after-match-do-while");
-
-  // Match while.
-  revng_log(BeautifyLogger, "Matching while\n");
-  matchWhile(RootNode, CombedAST);
-  Dumper.log("after-match-while");
-
   // Perform the `SwitchBreak` simplification
   revng_log(BeautifyLogger, "Performing SwitchBreak simplification");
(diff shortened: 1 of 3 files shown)


share/revng/test/tests/analysis/Decompilation/clang/dual-case-switch.c.model.yml
```diff
@@ -0,0 +1,33 @@
...
+    Arguments:
+      - Index: 0
+        Type:
+          UnqualifiedType: "/Types/520-PrimitiveType"
+  - ID: 1884
+    Kind: CABIFunctionType
+    ABI: SystemV_x86_64
+    ReturnType:
+      UnqualifiedType: "/Types/256-PrimitiveType"
+    Arguments: []
```
pfez: Is it really important to check the model here? Isn't the `FileCheck` enough? For instance, look at what we do for `pretty-ints`. In that case we only check the architecture, because the checks on the model aren't really important in that case. In principle I'm fine with having an empty YAML if we don't really care about checking the model. This makes the test more robust against changes in the model that are not significant for the feature being tested here. The only problem is that apparently right now `revng model compare` doesn't accept an empty YAML, so for all the tests where the model is not really important we just pass a YAML with only the architecture, which is basically guaranteed to pass unless something really bad happens.

andrealinux1: I agree with the fact that checking the model here is not particularly significant for the objectives of the test.

The point is that I saw that the current pipeline for the `revng-c.decompilation` tests type check again a known model with `revng model compare` as you mentioned, and I didn't want to craft another identical pipeline with the except of this `revng model compare` check, and at that point I thought of also including the checks against the names of the functions. But I agree that only including the check for the architecture may be more sensible. I'll dot it

andrealinux1: Ok, I left the architecture check only in https://github.com/revng/revng-c/commit/422d470fec9ff6e063489468d645c31e0a546808, for what concerns the model check.

PR #6, opened 2024-03-27
`switch` value simplification

dwarfyresearch:


lib/Canonicalize/CMakeLists.txt
```diff
@@ -2,6 +2,8 @@
 # This file is distributed under the MIT License. See LICENSE.md for details.
 #
 
+add_subdirectory(SimplifySwitch)
```
aleclearmind: 1. Remove subdirectory
2. Conflate the `*Pipe.cpp` and `*Pass.cpp` files
3. Drop the private header file

lib/Canonicalize/SimplifySwitch/SimplifySwitchPass.cpp
```diff
@@ -0,0 +1,234 @@
+//
+// Copyright rev.ng Labs Srl. See LICENSE.md for details.
```
aleclearmind: Adopt the new header

lib/Canonicalize/SimplifySwitch/SimplifySwitchPass.cpp
```diff
@@ -0,0 +1,234 @@
...
+#include "revng/Support/FunctionTags.h"
+#include "revng/Support/IRHelpers.h"
+#include "revng/ValueMaterializer/DataFlowGraph.h"
+#include "revng/ValueMaterializer/MemoryOracle.h"
+#include "revng/ValueMaterializer/ValueMaterializer.h"
+
+#include "revng-c/Pipes/Kinds.h"
+#include "revng-c/Support/FunctionTags.h"
+#include "revng-c/Support/IRHelpers.h"
+#include "revng-c/TypeNames/ModelTypeNames.h"
```
aleclearmind: Are these a minimal set of headers?

lib/Canonicalize/SimplifySwitch/SimplifySwitchPass.cpp
```diff
@@ -0,0 +1,234 @@
...
+
+#include "SimplifySwitchPass.h"
+
+using pipeline::serializedLocation;
+namespace ranks = revng::ranks;
+using BiDFNode = BidirectionalNode<DataFlowNode>;
+
+static Logger<> SwitchOptLogger("switch-opt");
```
aleclearmind: Rename to `Log`

lib/Canonicalize/SimplifySwitch/SimplifySwitchPass.cpp
```diff
@@ -0,0 +1,234 @@
...
+
+namespace {
```
aleclearmind: Why?

lib/Canonicalize/SimplifySwitch/SimplifySwitchPass.cpp
```diff
@@ -0,0 +1,234 @@
...
+class DataMemoryOracle final : public MemoryOracle {
+private:
+  RawBinaryView &BinaryView;
+  model::Architecture::Values Arch;
```
aleclearmind: 1. Initialize to `Invalid`.
2. No abbreviations.

lib/Canonicalize/SimplifySwitch/SimplifySwitchPass.cpp
```diff
@@ -0,0 +1,234 @@
...
+#include "SimplifySwitchPass.h"
+
+using pipeline::serializedLocation;
+namespace ranks = revng::ranks;
+using BiDFNode = BidirectionalNode<DataFlowNode>;
+
+static Logger<> SwitchOptLogger("switch-opt");
+
+namespace {
+class DataMemoryOracle final : public MemoryOracle {
```
aleclearmind: Rename to `RawBinaryMemoryOracle`

lib/Canonicalize/SimplifySwitch/SimplifySwitchPass.cpp
```diff
@@ -0,0 +1,234 @@
...
+public:
+  DataMemoryOracle(RawBinaryView &BinaryView,
+                   model::Architecture::Values Arch) :
+    BinaryView(BinaryView), Arch(Arch) {}
+  ~DataMemoryOracle() final = default;
+
+  MaterializedValue load(uint64_t LoadAddress, unsigned LoadSize) final {
+    auto Address = MetaAddress::fromGeneric(toLLVMArchitecture(Arch),
+                                            LoadAddress);
+    auto MaybeValue = BinaryView.readInteger(Address, LoadSize, IsLittleEndian);
```
aleclearmind: You're assuming big endian. Why?

Deduce endianess from `Arch`.

lib/Canonicalize/SimplifySwitch/SimplifySwitchPass.cpp
```diff
@@ -0,0 +1,234 @@
...
+private:
+  RawBinaryView &BinaryView;
+  model::Architecture::Values Arch;
+  bool IsLittleEndian = false;
+
+public:
+  DataMemoryOracle(RawBinaryView &BinaryView,
+                   model::Architecture::Values Arch) :
+    BinaryView(BinaryView), Arch(Arch) {}
+  ~DataMemoryOracle() final = default;
```
aleclearmind: Necessary?

lib/Canonicalize/SimplifySwitch/SimplifySwitchPass.cpp
```diff
@@ -0,0 +1,234 @@
...
+
+  MaterializedValue load(uint64_t LoadAddress, unsigned LoadSize) final {
+    auto Address = MetaAddress::fromGeneric(toLLVMArchitecture(Arch),
+                                            LoadAddress);
+    auto MaybeValue = BinaryView.readInteger(Address, LoadSize, IsLittleEndian);
+    auto NewAPInt = [LoadSize](uint64_t V) {
+      return llvm::APInt(LoadSize * 8, V);
+    };
```
aleclearmind: This does not seem necessary. Simplify.

lib/Canonicalize/SimplifySwitch/SimplifySwitchPass.cpp
```diff
@@ -0,0 +1,234 @@
...
+                                                       Value.getBitWidth());
+  if (DestinationType->isPointerTy()) {
+    revng_abort();
+  } else if (auto *IT = dyn_cast<IntegerType>(DestinationType)) {
+    revng_assert(IT->getBitWidth() == Value.getBitWidth());
+    return ConstantInt::get(DestinationType, Value);
+  } else {
+    revng_abort();
+  }
+}
```
aleclearmind: Do not duplicate code. Move the function to `IRHelpers.h` in `revng`.

lib/Canonicalize/SimplifySwitch/SimplifySwitchPass.cpp
```diff
@@ -0,0 +1,234 @@
...
+
+static llvm::BasicBlock *getBlockFor(llvm::SwitchInst *Switch,
+                                     llvm::Constant *C) {
+  uint64_t TheConstant = cast<llvm::ConstantInt>(C)->getValue().getZExtValue();
```
aleclearmind: Drop `->getValue().getZExtValue()`, below too. You can just compare the `Constant *`. If they are the same, the pointers will be the same too.

lib/Canonicalize/SimplifySwitch/SimplifySwitchPass.cpp
```diff
@@ -0,0 +1,234 @@
...
+
+  return nullptr;
+}
+
+static bool runSimplifySwitch(llvm::Function &F,
+                              llvm::LazyValueInfo &LVI,
+                              llvm::DominatorTree &DT,
+                              DataMemoryOracle &MO) {
+  if (F.isDeclaration() || !F.hasExactDefinition())
+    return false;
```
aleclearmind: 1. What does `hasExactDefinition` do?
2. Unnecessary.

lib/Canonicalize/SimplifySwitch/SimplifySwitchPass.cpp
```diff
@@ -0,0 +1,234 @@
...
+
+  for (auto &BB : F) {
```
aleclearmind: `auto` abuse, fix all the others too.
Also, never let `auto` infer `*`.

lib/Canonicalize/SimplifySwitch/SimplifySwitchPass.cpp
```diff
@@ -0,0 +1,234 @@
...
+      return I.getCaseSuccessor();
+  }
+
+  return nullptr;
+}
+
+static bool runSimplifySwitch(llvm::Function &F,
+                              llvm::LazyValueInfo &LVI,
+                              llvm::DominatorTree &DT,
+                              DataMemoryOracle &MO) {
```
aleclearmind: The way this function `return`s is broken.
If you seem to be assuming there's only one `switch` per function.

Consider having a function to handle a single `switch` so you can still do early returns.

lib/Canonicalize/SimplifySwitch/SimplifySwitchPass.cpp
```diff
@@ -0,0 +1,234 @@
...
+#include "revng-c/Pipes/Kinds.h"
+#include "revng-c/Support/FunctionTags.h"
+#include "revng-c/Support/IRHelpers.h"
+#include "revng-c/TypeNames/ModelTypeNames.h"
+
+#include "SimplifySwitchPass.h"
+
+using pipeline::serializedLocation;
+namespace ranks = revng::ranks;
+using BiDFNode = BidirectionalNode<DataFlowNode>;
```
aleclearmind: Bad name

lib/Canonicalize/SimplifySwitch/SimplifySwitchPass.cpp
```diff
@@ -0,0 +1,234 @@
...
+#include "revng/ValueMaterializer/ValueMaterializer.h"
+
+#include "revng-c/Pipes/Kinds.h"
+#include "revng-c/Support/FunctionTags.h"
+#include "revng-c/Support/IRHelpers.h"
+#include "revng-c/TypeNames/ModelTypeNames.h"
+
+#include "SimplifySwitchPass.h"
+
+using pipeline::serializedLocation;
```
aleclearmind: Unused

lib/Canonicalize/SimplifySwitch/SimplifySwitchPass.cpp
```diff
@@ -0,0 +1,234 @@
...
+namespace ranks = revng::ranks;
```
aleclearmind: Unused

lib/Canonicalize/SimplifySwitch/SimplifySwitchPass.cpp
```diff
@@ -0,0 +1,234 @@
...
+      if (ConditionInst and isa<llvm::PHINode>(ConditionInst))
+        continue;
+
+      if (StartNode->Value == Condition)
+        continue;
+
+      if (Results.values()->size() != Switch->getNumCases()) {
+        revng_log(SwitchOptLogger, "Unexpected values!");
+        continue;
+      }
```
aleclearmind: All these early continues are very mysterious.
Thoroughly comment all of them.

Also, why are we skipping `PHINode`s?

lib/Canonicalize/SimplifySwitch/SimplifySwitchPass.cpp
```diff
@@ -0,0 +1,234 @@
...
+                                                 &F);
+      llvm::IRBuilder<> IRBUnreachable(UnreachBB);
+      IRBUnreachable.CreateUnreachable();
+      auto *NewSwitch = Builder.CreateSwitch(StartNode->Value,
+                                             UnreachBB,
+                                             NewLabels.size());
+
+      for (auto &NewLabel : NewLabels) {
+        NewSwitch->addCase(NewLabel.first, NewLabel.second);
+      }
```
aleclearmind: 1. Confusing code organization.
2. Do not create an `IRBuilder` to emit a single instruction.

lib/Canonicalize/SimplifySwitch/SimplifySwitchPipe.cpp
```diff
@@ -0,0 +1,75 @@
...
+  PM.add(new LoadBinaryWrapperPass(Buffer->getBuffer()));
+  PM.add(new SimplifySwitchPass);
+
+  PM.run(TargetsList.getModule());
+}
+
+llvm::Error
+SimplifySwitch::checkPrecondition(const pipeline::Context &Ctx) const {
+  return llvm::Error::success();
+}
```
aleclearmind: Inline

lib/Canonicalize/SimplifySwitch/SimplifySwitchPass.cpp
```diff
@@ -0,0 +1,234 @@
+//
+// Copyright rev.ng Labs Srl. See LICENSE.md for details.
+//
+
```
aleclearmind: Add detailed comment describing this pass

lib/Canonicalize/SimplifySwitch/SimplifySwitchPipe.cpp
```diff
@@ -0,0 +1,75 @@
...
+
+namespace revng::pipes {
+
+class SimplifySwitch {
+public:
+  static constexpr auto Name = "simplify-switch";
+
+  std::array<pipeline::ContractGroup, 1> getContract() const {
+    // // TODO: Is an empty contract valid?
+    return { pipeline::ContractGroup({}) };
```
aleclearmind: 1. Don't use `TODO` for things that are broken, use `WIP` so we're sure this doesn't go in.
2. @drblallo this pipe reads the model, the binary and makes some changes to the IR. How should the contract look like?

aleclearmind (changes requested):

lib/Canonicalize/SimplifySwitch/SimplifySwitchPass.cpp
dwarfyresearch: No.

lib/Canonicalize/SimplifySwitch/SimplifySwitchPass.cpp
dwarfyresearch: Not needed.

lib/Canonicalize/SimplifySwitch/SimplifySwitchPass.cpp
dwarfyresearch: No.

lib/Canonicalize/SimplifySwitch/SimplifySwitchPass.cpp
dwarfyresearch: Handled by using `bool Result`.

lib/Canonicalize/SimplifySwitch/SimplifySwitchPass.cpp
dwarfyresearch: Most of them are removed now.

We were skipping phis as a workaround, but it is not needed anymore.

lib/Canonicalize/SimplifySwitch/SimplifySwitchPass.cpp
dwarfyresearch: We usually skip those when writing LLVM Passes since some functions can have `derefined` definitions, but yeah, it does not apply to this use case too much.

share/revng/pipelines/revng-c-pipelines.yml
```diff
@@ -63,6 +63,23 @@ Branches:
               - instsimplify
               - dse
               - dce
+      - Name: simplify-switch
+        Pipes:
+          - Type: simplify-switch
+            UsedContainers: [input, module.ll]
+      - Name: post-swtich-simplify-cleanup
```
aleclearmind: 1. Typo (`swtich`)
2. Why a new step? Seem unnecessary.
3. Why this pipeline? How did you put it together? What happens if we don't run any of these?

lib/Canonicalize/SimplifySwitchPass.cpp
```diff
@@ -0,0 +1,263 @@
...
+    for (llvm::Instruction &I : BB) {
+      llvm::SwitchInst *Switch = dyn_cast<llvm::SwitchInst>(&I);
+      if (not Switch)
+        continue;
+
+      // Skip empty switches.
+      if (Switch->getNumCases() == 0)
+        continue;
+
+      auto *Condition = Switch->getCondition();
```
aleclearmind: Fix all the auto abuses

lib/Canonicalize/SimplifySwitchPass.cpp
```diff
@@ -0,0 +1,263 @@
...
+  bool Result = false;
+  llvm::SmallVector<llvm::SwitchInst *, 16> SwitchToDelete;
+  for (llvm::BasicBlock &BB : F) {
+    for (llvm::Instruction &I : BB) {
+      llvm::SwitchInst *Switch = dyn_cast<llvm::SwitchInst>(&I);
+      if (not Switch)
+        continue;
+
+      // Skip empty switches.
+      if (Switch->getNumCases() == 0)
```
aleclearmind: Outline the body of this loop into a function taking a `SwitchInst *` and replace  `continue`s with `return`s.

lib/Canonicalize/SimplifySwitchPass.cpp
```diff
@@ -0,0 +1,263 @@
...
+            BidirectionalNode<DataFlowNode>
+              *N = const_cast<BidirectionalNode<DataFlowNode> *>(BidirectNode);
+            StartNode = N;
+          }
+        }
+      }
+
+      // We found more than one start node.
+      if (StartNode == nullptr or BailOut)
+        continue;
```
aleclearmind: Outline identifying `StartNode` into a function.

lib/Canonicalize/SimplifySwitchPass.cpp
```diff
@@ -0,0 +1,263 @@
...
+
+      std::map<llvm::ConstantInt *, llvm::BasicBlock *> NewLabels;
+      for (auto &Value : *StartNode->OracleRange) {
+        auto Old = DFG.materializeOne(StartNode, MO, Value);
+        if (not Old)
+          continue;
```
aleclearmind: 1. This and the next `continue` work, but if you outline the body of the `for` into a function you can turn these into a `return`, which is better.
2. For each failing case, print a diagnosis to the logger.

lib/Canonicalize/SimplifySwitchPass.cpp
```diff
@@ -0,0 +1,263 @@
...
+#include "llvm/Passes/PassBuilder.h"
+
+#include "revng/Lift/LoadBinaryPass.h"
+#include "revng/Pipeline/RegisterPipe.h"
+#include "revng/Pipes/FileContainer.h"
+#include "revng/ValueMaterializer/ValueMaterializer.h"
+
+#include "revng-c/Support/IRHelpers.h"
+
+static Logger<> Log("switch-opt");
```
aleclearmind, line 30: Add `using namespace llvm` and drop all the `llvm::`.

lib/Canonicalize/SimplifySwitchPass.cpp
```diff
@@ -0,0 +1,263 @@
...
+          continue;
+        NewLabels[cast<llvm::ConstantInt>(ConstantForTheValue)] = BlockForValue;
+      }
+
+      // We have not found all the cases for the new switch.
+      if (Switch->getNumCases() != NewLabels.size())
+        continue;
+
+      // Recreate the new switch.
+      llvm::SwitchInst
```
aleclearmind: Here you can use `auto *`.

share/revng/test/tests/analysis/CollectCFG/x86_64/switch-jump-table.S.filecheck
```diff
@@ -0,0 +1,15 @@
...
+CHECK: %1 = inttoptr i64 %rax_x86_64 to ptr
+CHECK: %2 = load i64, ptr %1, align 1
+CHECK: %3 = icmp ugt i64 %2, 5
+CHECK: switch i64 %2
+CHECK: i64 {{[0-5]+}}, label
+CHECK: i64 {{[0-5]+}}, label
+CHECK: i64 {{[0-5]+}}, label
+CHECK: i64 {{[0-5]+}}, label
+CHECK: i64 {{[0-5]+}}, label
+CHECK: i64 {{[0-5]+}}, label
```
aleclearmind: I think you can hardcode these, no need for a regular expression.
In fact, this is the only thing I would test: let's check that there's a `switch` statement with the exact number of case and values we expect (and not more/less).
Testing the rest is just making the test more fragile, for instance, we're going to drop `_myfunc` from the function name in favor of the `MetaAddress` of the function.

aleclearmind (changes requested):

lib/Canonicalize/SimplifySwitch/SimplifySwitchPipe.cpp
aleclearmind: @dwarfyresearch [here's an example](https://github.com/revng/revng-c/blob/9624f2bbd14b6a3aead30f7c96fc6ed9a0b255ad/lib/RemoveLiftingArtifacts/MakeSegmentRefPipe.cpp#L26C1-L33C4) of what it should look like. Make sure the kind of the functions is the correct one.

share/revng/pipelines/revng-c-pipelines.yml
dwarfyresearch: It was something that I needed to workaround some bug in my impl., but it is not needed anymore.

share/revng/test/tests/analysis/CollectCFG/x86_64/switch-jump-table.S.filecheck
dwarfyresearch: I see OK. I will test the switch only. In general, I'm using the regexp since I cannot rely on the order of the cases.

aleclearmind: You can use `CHECK-DAG` then I think

lib/Canonicalize/SimplifySwitchPass.cpp
```diff
@@ -0,0 +1,263 @@
...
+
+namespace revng::pipes {
+
+class SimplifySwitch {
+public:
+  static constexpr auto Name = "simplify-switch";
+
+  std::array<pipeline::ContractGroup, 1> getContract() const {
+    // // WIP: Is an empty contract valid?
+    return { pipeline::ContractGroup({}) };
```
aleclearmind: This needs to be fixed

lib/Canonicalize/SimplifySwitchPass.cpp
```diff
@@ -0,0 +1,263 @@
...
+    return llvm::Error::success();
+  }
+
+  void print(const pipeline::Context &Ctx,
+             llvm::raw_ostream &OS,
+             llvm::ArrayRef<std::string> ContainerNames) const {
+    revng_check(ContainerNames.size() == 2);
+    OS << *ResourceFinder.findFile("bin/revng");
+    OS << " opt SimplifySwitch " << ContainerNames[0] << " "
+       << ContainerNames[1] << "\n";
```
aleclearmind: Leave this method empty, it's deprecated

share/revng/pipelines/revng-c-pipelines.yml
aleclearmind: This is not fixed

share/revng/test/configuration/revng-c/simplify-switch.yml
```diff
@@ -0,0 +1,16 @@
...
+  # Run simplify-switch and test against ground truth
+  #
+  - type: revng-c.test-simplify-switch
+    from:
+      - type: revng-qa.compiled-with-debug-info
+        filter: for-simplify-switch
+    suffix: /
+    command: |-
+      revng artifact --resume "$OUTPUT" decompile-to-single-file --abi-enforcement-level=no --analyses-list=revng-initial-auto-analysis --analyses-list=revng-c-initial-auto-analysis "$INPUT" -o /dev/null;
+      FileCheck --input-file="$OUTPUT/simplify-switch/module.ll" "${SOURCE}.filecheck";
```
aleclearmind: 1. Split this over multiple lines (no `\` at the end of line required)
2. Do not use `--analyses-list=revng-initial-auto-analysis --analyses-list=revng-c-initial-auto-analysis`, just use `--analyze`
3. There's no reason to get to `decompile-to-single-file`. Just do `revng artifact simplify-switch | FileCheck "${SOURCE}.filecheck"`

share/revng/test/tests/analysis/CollectCFG/x86_64/switch-jump-table.S.filecheck
aleclearmind: Not resolved

aleclearmind (changes requested):

lib/Canonicalize/SimplifySwitch/SimplifySwitchPass.cpp
aleclearmind: Review all the others files

lib/Canonicalize/SimplifySwitchPass.cpp
aleclearmind: Please outline the loop, it should improve readability

lib/Canonicalize/SimplifySwitchPass.cpp
aleclearmind: Not done

lib/Canonicalize/SimplifySwitchPass.cpp
aleclearmind: Not fixed

lib/Canonicalize/SimplifySwitchPass.cpp
aleclearmind: Not fixed

lib/Canonicalize/SimplifySwitchPass.cpp
aleclearmind: Not done

lib/Canonicalize/SimplifySwitchPass.cpp
dwarfyresearch: I have not pushed the code back then, but I marked all the comments as resolved...

share/revng/test/configuration/revng-c/simplify-switch.yml
dwarfyresearch: Addressed most of the items (except `3`). Running `revng artifact simplify-switch` on this binary triggers an unrelated assert, so that is why I still use `decompile-to-single-file` (even some other artifacts do not work, e.g. `$ revng artifact --resume workdir/ remove-lifting-artifacts switch-jump-table-revng-qa.compiled-with-debug-info-6f2b090f  -o /dev/null ;` triggers an assert).

lib/Canonicalize/SimplifySwitchPass.cpp
```diff
@@ -0,0 +1,282 @@
...
+
+namespace revng::pipes {
+
+class SimplifySwitch {
+public:
+  static constexpr auto Name = "simplify-switch";
+
+  std::array<pipeline::ContractGroup, 1> getContract() const {
+    pipeline::Contract BinaryPart(kinds::Binary, 0, kinds::Binary, 0);
+    pipeline::Contract FunctionsPart(kinds::Isolated, 1);
```
aleclearmind: You need to use the kind produced by the last pipe, i.e., `kinds::StackPointerPromoted,`.

share/revng/test/configuration/revng-c/simplify-switch.yml
aleclearmind: Please report the exact problem on mattermost, this needs to be fixed.

I doubt that `decompile-to-single-file` succeeds when `simplify-switch` doesn't. However it's not impossible givent that the contract for the pipe is incorrect. Try to fix it as suggested and try again.

dwarfyresearch: Done.

pfez:
Merged in 6d071d0a5da3c07de35cde51a9f946ed210cf0e0

Thanks!

PR #9, opened 2024-04-15
Improve `InlineDispatcherSwitch` logic

andrealinux1:
We introduce the following improvements for the `InlineDispatcherSwitch` routine:
- We do not inline `case`s which contain a `break` or `continue` statements. Moving such kind of statement would indeed break the semantics.
- Simplify redundant _non local control flow_ statements: i.e., when we inline a `return` statement, we remove redundant following `break`s and `continue`s.
- We remove the associated `SetNode`s when simplifying `SwitchBreak` statements for exit dispatchers.

Additionally, we remove some dead code and reorder the `SwitchBreaksFixer` pass.

### DIFF
--- a/include/revng-c/RestructureCFG/ASTNode.h
+++ b/include/revng-c/RestructureCFG/ASTNode.h
@@ -405,6 +405,8 @@ public:
 
   links_container::size_type length() const { return NodeVec.size(); }
 
+  bool empty() const { return NodeVec.empty(); }
+
   ASTNode *getNodeN(links_container::size_type N) const { return NodeVec[N]; }
 
   links_container &getChildVec() { return NodeVec; }
--- a/lib/RestructureCFG/BeautifyGHAST.cpp
+++ b/lib/RestructureCFG/BeautifyGHAST.cpp
@@ -1016,12 +1016,12 @@ void beautifyAST(const model::Binary &Model, Function &F, ASTTree &CombedAST) {
 
   // Perform the `SwitchBreak` simplification
   revng_log(BeautifyLogger, "Performing SwitchBreak simplification");
-  RootNode = simplifySwitchBreak(CombedAST, RootNode);
+  RootNode = simplifySwitchBreak(CombedAST);
   Dumper.log("After-switchbreak-simplify");
 
   // Perform the dispatcher `switch` inlining
   revng_log(BeautifyLogger, "Performing dispatcher switch inlining\n");
-  RootNode = inlineDispatcherSwitch(CombedAST, RootNode);
+  RootNode = inlineDispatcherSwitch(CombedAST);
   Dumper.log("after-dispatcher-switch-inlining");
 
   // Perform the simplification of `switch` with two entries in a `if`
--- a/lib/RestructureCFG/BeautifyGHAST.cpp
+++ b/lib/RestructureCFG/BeautifyGHAST.cpp
@@ -28,6 +28,7 @@
 #include "FallThroughScopeAnalysis.h"
 #include "InlineDispatcherSwitch.h"
 #include "PromoteCallNoReturn.h"
+#include "RemoveDeadCode.h"
 #include "SimplifyCompareNode.h"
 #include "SimplifyDualSwitch.h"
 #include "SimplifyHybridNot.h"
@@ -1023,6 +1024,14 @@ void beautifyAST(const model::Binary &Model, Function &F, ASTTree &CombedAST) {
   revng_log(BeautifyLogger, "Performing dispatcher switch inlining\n");
   RootNode = inlineDispatcherSwitch(CombedAST);
   Dumper.log("after-dispatcher-switch-inlining");
+
+  // Perform the dead code simplification.
+  // We invoke this pass here because the dispatcher case inlining may have
+  // moved around some non local control flow statements like `return`, in such
+  // a way that a dead code simplification step is needed.
+  revng_log(BeautifyLogger, "Performing dead code simplification\n");
+  RootNode = removeDeadCode(Model, CombedAST);
+  Dumper.log("after-dead-code-simplify");
 
   // Perform the simplification of `switch` with two entries in a `if`
   revng_log(BeautifyLogger, "Performing the dual switch simplification\n");
--- a/lib/RestructureCFG/BeautifyGHAST.cpp
+++ b/lib/RestructureCFG/BeautifyGHAST.cpp
@@ -1029,11 +1029,6 @@ void beautifyAST(const model::Binary &Model, Function &F, ASTTree &CombedAST) {
   RootNode = simplifyDualSwitch(CombedAST, RootNode);
   Dumper.log("after-dual-switch-simplify");
 
-  // Fix loop breaks from within switches
-  revng_log(BeautifyLogger, "Fixing loop breaks inside switches\n");
-  SwitchBreaksFixer().run(RootNode, CombedAST);
-  Dumper.log("after-fix-switch-breaks");
-
   // Remove empty sequences.
   revng_log(BeautifyLogger, "Removing empty sequence nodes\n");
   RootNode = simplifyAtomicSequence(CombedAST, RootNode);
@@ -1090,6 +1085,11 @@ void beautifyAST(const model::Binary &Model, Function &F, ASTTree &CombedAST) {
   revng_log(BeautifyLogger, "Performing the implicit return simplification\n");
   simplifyImplicitReturn(CombedAST, RootNode);
   Dumper.log("after-implicit-return-simplify");
+
+  // Fix loop breaks from within switches
+  revng_log(BeautifyLogger, "Fixing loop breaks inside switches\n");
+  SwitchBreaksFixer().run(RootNode, CombedAST);
+  Dumper.log("after-fix-switch-breaks");
 
   // Serialize the collected metrics in the statistics file if necessary
   if (StatsFileStream) {
--- a/lib/RestructureCFG/BeautifyGHAST.cpp
+++ b/lib/RestructureCFG/BeautifyGHAST.cpp
@@ -33,7 +33,6 @@
 #include "SimplifyHybridNot.h"
 #include "SimplifyImplicitStatement.h"
 
-using std::make_unique;
 using std::unique_ptr;
 
 using namespace llvm;
--- a/lib/RestructureCFG/InlineDispatcherSwitch.cpp
+++ b/lib/RestructureCFG/InlineDispatcherSwitch.cpp
@@ -221,6 +221,15 @@ static RecursiveCoroutine<ASTNode *> addToDispatcherSet(ASTTree &AST,
   rc_return Node;
 }
 
+/// Wrapper helper used to remove a specific `SetNode` from the body of a loop,
+/// reusing the code of the `addToDispatcherSet` helper
+static void removeDispatcherSet(ASTTree &AST, ASTNode *Node, ASTNode *Set) {
+
+  // We perform the `SetNode` by instructing `addToDispatcherSet` to inline a
+  // `nullptr`, and to remove the original `SetNode`
+  addToDispatcherSet(AST, Node, nullptr, Set, true);
+}
+
 static bool isDispatcherIf(ASTNode *If) {
   const ExprNode *E = llvm::cast<IfNode>(If)->getCondExpr();
   using NodeKind = ExprNode::NodeKind;
@@ -771,7 +780,10 @@ ASTNode *inlineDispatcherSwitch(ASTTree &AST, ASTNode *RootNode) {
   return RootNode;
 }
 
-static RecursiveCoroutine<ASTNode *> simplifySwitchBreakImpl(ASTNode *Node) {
+static RecursiveCoroutine<ASTNode *>
+simplifySwitchBreakImpl(ASTTree &AST,
+                        ASTNode *Node,
+                        const LoopDispatcherMap &LoopDispatcherMap) {
   switch (Node->getKind()) {
   case ASTNode::NK_List: {
     SequenceNode *Seq = llvm::cast<SequenceNode>(Node);
@@ -779,7 +791,7 @@ static RecursiveCoroutine<ASTNode *> simplifySwitchBreakImpl(ASTNode *Node) {
     // In place of a sequence node, we just need to inspect all the nodes in the
     // sequence
     for (ASTNode *&N : Seq->nodes()) {
-      N = rc_recur simplifySwitchBreakImpl(N);
+      N = rc_recur simplifySwitchBreakImpl(AST, N, LoopDispatcherMap);
     }
 
     // In this beautify, it may be that a dispatcher switch is completely
@@ -795,7 +807,9 @@ static RecursiveCoroutine<ASTNode *> simplifySwitchBreakImpl(ASTNode *Node) {
     // Inspect loop nodes
     if (Scs->hasBody()) {
       ASTNode *Body = Scs->getBody();
-      ASTNode *NewBody = rc_recur simplifySwitchBreakImpl(Body);
+      ASTNode *NewBody = rc_recur simplifySwitchBreakImpl(AST,
+                                                          Body,
+                                                          LoopDispatcherMap);
       Scs->setBody(NewBody);
     }
 
@@ -806,12 +820,16 @@ static RecursiveCoroutine<ASTNode *> simplifySwitchBreakImpl(ASTNode *Node) {
     // Inspect the `then` and `else` branches
     if (If->hasThen()) {
       ASTNode *Then = If->getThen();
-      ASTNode *NewThen = rc_recur simplifySwitchBreakImpl(Then);
+      ASTNode *NewThen = rc_recur simplifySwitchBreakImpl(AST,
+                                                          Then,
+                                                          LoopDispatcherMap);
       If->setThen(NewThen);
     }
     if (If->hasElse()) {
       ASTNode *Else = If->getElse();
-      ASTNode *NewElse = rc_recur simplifySwitchBreakImpl(Else);
+      ASTNode *NewElse = rc_recur simplifySwitchBreakImpl(AST,
+                                                          Else,
+                                                          LoopDispatcherMap);
       If->setElse(NewElse);
     }
   } break;
@@ -824,8 +842,8 @@ static RecursiveCoroutine<ASTNode *> simplifySwitchBreakImpl(ASTNode *Node) {
     for (auto &Group : llvm::enumerate(Switch->cases())) {
       unsigned Index = Group.index();
       auto &LabelCasePair = Group.value();
-      LabelCasePair
-        .second = rc_recur simplifySwitchBreakImpl(LabelCasePair.second);
+      LabelCasePair.second = rc_recur
+        simplifySwitchBreakImpl(AST, LabelCasePair.second, LoopDispatcherMap);
 
       if (LabelCasePair.second == nullptr) {
         ToRemoveCaseIndex.push_back(Index);
@@ -854,6 +872,9 @@ static RecursiveCoroutine<ASTNode *> simplifySwitchBreakImpl(ASTNode *Node) {
       // Search for cases that are composed by a single `SwitchBreak` node
       llvm::SmallVector<size_t> ToRemoveCaseIndex;
 
+      // Save the `LabelSet`s associated to the removed `case`s
+      llvm::SmallSet<size_t, 1> RemovedLabels;
+
       // We do not need to skip the `default` case here, because there is no
       // `default` in first place
       for (auto &Group : llvm::enumerate(Switch->cases())) {
@@ -862,11 +883,41 @@ static RecursiveCoroutine<ASTNode *> simplifySwitchBreakImpl(ASTNode *Node) {
 
         if (llvm::isa<SwitchBreakNode>(Case)) {
           ToRemoveCaseIndex.push_back(Index);
+
+          // If we are removing an atomic `Label`, we save it for later removal
+          // of the associated `SetNode`. We do not remove the `SetNode`s
+          // associated to multiple `Label`s `case`s, because that would not be
+          // correct. The removal of the associated `SetNode` should be
+          // performed only in correspondence of the `case`s containing the
+          // atomic `Label`.
+          if (LabelSet.size() == 1) {
+            RemovedLabels.insert(*LabelSet.begin());
+          }
         }
       }
 
       for (auto ToRemoveCase : llvm::reverse(ToRemoveCaseIndex)) {
         Switch->removeCaseN(ToRemoveCase);
+      }
+
+      // In case we are simplifying the `SwitchBreak` associated to an exit
+      // dispatcher, we also need to remove the corresponding `SetNode`s
+      using DispatcherKind = typename SwitchNode::DispatcherKind;
+      if (not Switch->getCondition()
+          and Switch->getDispatcherKind() == DispatcherKind::DK_Exit) {
+
+        // Precompute how many `SetNode`s are present in the related `Scs`
+        auto &[RelatedLoop, RemoveSetNode] = LoopDispatcherMap.at(Switch);
+        SetNodeCounterMap SetCounterMap;
+        countSetNodeInLoop(RelatedLoop->getBody(), SetCounterMap);
+
+        // Remove the `SetNode`s associated to the removed `Label`s
+        for (auto &Label : RemovedLabels) {
+          auto &Sets = SetCounterMap.at(Label);
+          for (auto *Set : Sets) {
+            removeDispatcherSet(AST, RelatedLoop->getBody(), Set);
+          }
+        }
       }
     }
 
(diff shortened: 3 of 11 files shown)


include/revng-c/RestructureCFG/ASTNode.h
```diff
@@ -405,6 +405,8 @@ class SequenceNode : public ASTNode {
 
   links_container::size_type length() const { return NodeVec.size(); }
 
+  bool isEmpty() const { return length() == 0; }
```
pfez: Just `empty()` maybe? It's more consistent with STL practice.

lib/RestructureCFG/InlineDispatcherSwitch.cpp
```diff
@@ -509,6 +509,65 @@ static void processNestedWeavedSwitches(SwitchNode *Switch) {
...
 
+static RecursiveCoroutine<bool> containsNonLocalCF(ASTNode *Node) {
+  switch (Node->getKind()) {
+  case ASTNode::NK_List: {
+    SequenceNode *Seq = llvm::cast<SequenceNode>(Node);
+
+    bool HasNonLocalCF = false;
+    for (ASTNode *N : Seq->nodes()) {
+      HasNonLocalCF = HasNonLocalCF or rc_recur containsNonLocalCF(N);
+    }
```
pfez: Wouldn't it be faster to iterate in reverse order?
The very moment `HasNonLocalcCF` is set to true, you can just bail out.

lib/RestructureCFG/InlineDispatcherSwitch.cpp
```diff
@@ -509,6 +509,65 @@ static void processNestedWeavedSwitches(SwitchNode *Switch) {
...
+
+    rc_return HasNonLocalCF;
+  } break;
+  case ASTNode::NK_Scs: {
+
+    // We do not want to move around a `ScsNode` for any reason
```
pfez: We don't because we already know it creates problems? Or it's just an advanced scenario we haven't fully considered yet?

lib/RestructureCFG/InlineDispatcherSwitch.cpp
```diff
@@ -509,6 +509,65 @@ static void processNestedWeavedSwitches(SwitchNode *Switch) {
...
+
+    bool HasNonLocalCF = false;
+    if (If->hasThen()) {
+      ASTNode *Then = If->getThen();
+      HasNonLocalCF = HasNonLocalCF or rc_recur containsNonLocalCF(Then);
+    }
+    if (If->hasElse()) {
+      ASTNode *Else = If->getElse();
+      HasNonLocalCF = HasNonLocalCF or rc_recur containsNonLocalCF(Else);
+    }
```
pfez: Change this code so that it early returns `true` when possible.

Like, if the `then` has non-local control flow, there's no need to look inside the `else`.

lib/RestructureCFG/InlineDispatcherSwitch.cpp
```diff
@@ -509,6 +509,65 @@ static void processNestedWeavedSwitches(SwitchNode *Switch) {
...
+    rc_return HasNonLocalCF;
+  } break;
+  case ASTNode::NK_Switch: {
+    auto *Switch = llvm::cast<SwitchNode>(Node);
+
+    bool HasNonLocalCF = false;
+    for (auto &LabelCasePair : Switch->cases()) {
+      HasNonLocalCF = HasNonLocalCF
+                      or rc_recur containsNonLocalCF(LabelCasePair.second);
+    }
```
pfez: Same as above, you can early exit at the first case that has non-local cf.

lib/RestructureCFG/InlineDispatcherSwitch.cpp
```diff
@@ -509,6 +509,65 @@ static void processNestedWeavedSwitches(SwitchNode *Switch) {
...
+
+    rc_return HasNonLocalCF;
+  } break;
+  case ASTNode::NK_Set:
+  case ASTNode::NK_Code:
+  case ASTNode::NK_SwitchBreak: {
+    rc_return false;
+  } break;
```
pfez: I'm not 100% sure about the `SwitchBreak`. Can you please elaborate?

lib/RestructureCFG/InlineDispatcherSwitch.cpp
```diff
@@ -737,13 +737,140 @@ inlineDispatcherSwitchImpl(ASTTree &AST,
...
+                                ASTTree &AST,
+                                ASTNode *RootNode) {
 
   // Compute the loop -> dispatcher switch correspondence
   LoopDispatcherMap LoopDispatcherMap = computeLoopDispatcher(AST);
 
   RootNode = inlineDispatcherSwitchImpl(AST, RootNode, LoopDispatcherMap);
 
+  // Pre-compute the `FallThroughScopeType` before the `SuperfluousNonLocalCF`
+  // simplidy is performed. This information is useful in order to decide
```
pfez: Typo: simpliDy

lib/RestructureCFG/InlineDispatcherSwitch.cpp
```diff
@@ -737,13 +737,140 @@ inlineDispatcherSwitchImpl(ASTTree &AST,
...
+  // Call the underlying implementation
+  RootNode = simplifySuperfluousNonLocalCFImpl(RootNode, FallThroughScopeMap);
+  AST.setRoot(RootNode);
+
+  return RootNode;
+}
+
+ASTNode *inlineDispatcherSwitch(const model::Binary &Model,
+                                ASTTree &AST,
+                                ASTNode *RootNode) {
```
pfez: At this point I don't think you even need to take the `RootNode` argument explicitly.
It seems to me that the whole code always assumes that `RootNode == AST.getRoot()`. Isn't that the case?
If I'm right, passing `RootNode` explicitly opens up opportunities for callers to pass a wrong `RootNode`, so I'd just remove it.

Same holds for `inlineDispatcherSwitch` below. Please fix that too.

lib/RestructureCFG/InlineDispatcherSwitch.cpp
```diff
@@ -737,13 +737,140 @@ inlineDispatcherSwitchImpl(ASTTree &AST,
...
+    // sequence
+    bool ShouldSimplify = false;
+    for (ASTNode *&N : Seq->nodes()) {
+      if (not ShouldSimplify) {
+
+        // Compute the `FallThrough information after each `SequencenNode` is
+        // processed
+        N = rc_recur simplifySuperfluousNonLocalCFImpl(N, FallThroughScopeMap);
+        FallThroughScopeType ReturnType = FallThroughScopeType::Return;
+        ShouldSimplify = (FallThroughScopeMap.at(N) == ReturnType);
```
pfez: Why the parentheses on RHS? Please drop them.

lib/RestructureCFG/InlineDispatcherSwitch.cpp
```diff
@@ -737,13 +737,140 @@ inlineDispatcherSwitchImpl(ASTTree &AST,
...
+      } else {
+        N = nullptr;
+      }
+    }
+
+    // In this beautify, it may be that a dispatcher switch is completely
+    // removed, therefore leaving a `nullptr` in a `SequenceNode`. We remove all
+    // these `nullptr` from the `SequenceNode` after the processing has taken
+    // place
+    Seq->removeNode(nullptr);
```
pfez: Is this guaranteed to always remove all of the `nullptr`s left? Or only the first one?

lib/RestructureCFG/InlineDispatcherSwitch.cpp
```diff
@@ -737,13 +737,140 @@ inlineDispatcherSwitchImpl(ASTTree &AST,
...
+
+    // This simplify should never completely empty a `SequenceNode`
+    revng_assert(not Seq->isEmpty());
```
pfez: So, worst case, the sequence is left with a single node.

Don't we have an invariant that requires sequence composed of only 1 node to be unwrapped?

lib/RestructureCFG/InlineDispatcherSwitch.cpp
```diff
@@ -968,12 +980,44 @@ static RecursiveCoroutine<ASTNode *> simplifySwitchBreakImpl(ASTNode *Node) {
...
+        SetNodeCounterMap SetCounterMap;
+        countSetNodeInLoop(RelatedLoop->getBody(), SetCounterMap);
+
+        // Remove the `SetNode`s associated to the removed `Label`s
+        for (auto &Label : RemovedAtomicLabels) {
+          auto &Sets = SetCounterMap.at(Label);
+          for (auto *Set : Sets) {
+
+            // To perform the removal, we reuse the `addToDispatcherSet`
+            addToDispatcherSet(AST, RelatedLoop->getBody(), nullptr, Set, true);
```
pfez: Please make a dedicated directive for this. Using `addToDispatcherSet` with `nullptr` leaks internal details and is slightly confusing if you ignore those details.

lib/RestructureCFG/InlineDispatcherSwitch.cpp
```diff
@@ -1002,8 +1046,11 @@ static RecursiveCoroutine<ASTNode *> simplifySwitchBreakImpl(ASTNode *Node) {
 /// `case`.
 ASTNode *simplifySwitchBreak(ASTTree &AST, ASTNode *RootNode) {
```
pfez: Here too I think we don't need the `RootNode` argument. All the code assumes `RootNode` is the root of `AST`, so `AST` is enough.

lib/RestructureCFG/InlineDispatcherSwitch.cpp
```diff
@@ -960,6 +969,9 @@ static RecursiveCoroutine<ASTNode *> simplifySwitchBreakImpl(ASTNode *Node) {
       // Search for cases that are composed by a single `SwitchBreak` node
       llvm::SmallVector<size_t> ToRemoveCaseIndex;
 
+      // Save the `LabelSet`s associated to the removed `case`s
+      llvm::SmallSet<size_t, 1> RemovedAtomicLabels;
```
pfez: Why `Atomic`?

pfez: Does that just mean that they have a single label and not many? If so, drop `Atomic` from the name, it seems to hint at some atomicity of the execution or something, which isn't the case.

lib/RestructureCFG/InlineDispatcherSwitch.cpp
```diff
@@ -737,13 +737,140 @@ inlineDispatcherSwitchImpl(ASTTree &AST,
...
+  // Pre-compute the `FallThroughScopeType` before the `SuperfluousNonLocalCF`
+  // simplidy is performed. This information is useful in order to decide
+  // whether we remove some statements that after the `case` inlining are
+  // preceded by `return` statements.
+  FallThroughScopeTypeMap
+    FallThroughScopeMap = computeFallThroughScope(Model, RootNode);
+
+  // Perform the `SuperfluousNonLocalCF` simplification pass
+  RootNode = simplifySuperfluousNonLocalCF(AST, RootNode, FallThroughScopeMap);
+
```
pfez: This looks like a dead-code-elimination pass.
Please carve this out in a separate beautifier.
And give it a name like `removeDeadCode`.

lib/RestructureCFG/InlineDispatcherSwitch.cpp
```diff
@@ -737,13 +737,140 @@ inlineDispatcherSwitchImpl(ASTTree &AST,
...
+    // In place of a sequence node, we just need to inspect all the nodes in the
+    // sequence
+    bool ShouldSimplify = false;
+    for (ASTNode *&N : Seq->nodes()) {
+      if (not ShouldSimplify) {
+
+        // Compute the `FallThrough information after each `SequencenNode` is
+        // processed
+        N = rc_recur simplifySuperfluousNonLocalCFImpl(N, FallThroughScopeMap);
+        FallThroughScopeType ReturnType = FallThroughScopeType::Return;
```
pfez: We should do something about calls to noreturn functions too.

lib/RestructureCFG/InlineDispatcherSwitch.cpp
andrealinux1: Yes, the `removeNode` method of  `SequenceNode` takes care of removing all the instances of empty nodes

lib/RestructureCFG/InlineDispatcherSwitch.cpp
andrealinux1: You are right, but there are some caveats that we informally discussed in the past, briefly:
- In principle, we could run the three main steps to enforce all the invariants at each beautify step, by invoking the `normalize` method defined [here](https://github.com/revng/revng-c/blob/9624f2bbd14b6a3aead30f7c96fc6ed9a0b255ad/include/revng-c/RestructureCFG/GenerateAst.h#L1005), but this would mean traversing the AST multiple times.
- Therefore, we decided in the past to keep the passes that enforce a single invariant separately available.
-  And this is the case we are in. If you take a look a the beautify passes "pipeline" [here](https://github.com/revng/revng-c/blob/9624f2bbd14b6a3aead30f7c96fc6ed9a0b255ad/lib/RestructureCFG/BeautifyGHAST.cpp#L1025), you can see that the `simplifyAtomicSequence` pass (which is in charge to unwrap unitary sequences), is run after `inlineDispatcherSwitch`, therefore eventually enforcing the property you mentioned, with the assurance that with the current anatomy of the "pipeline" we do not have any other pass that requires such property enforced, between `inlineDispatcherSwitch` and `simplifyAtomicSequence` itself.

- We could, o.c., run it just after `inlineDispatcherSwitch` , and this would prevent possible problems with future reordering of the steps of the "pipeline", at the cost of visiting more times the AST. I think it could be worth discussing about this IRL if you want more details.

lib/RestructureCFG/InlineDispatcherSwitch.cpp
andrealinux1: The reasoning is the following:
- A `SwitchBreak` can appear in the AST as the single element composing the body of a `case` of a `switch` node. Its meaning is equivalent to `nop` (e.g., a certain state variable assignment of an exit dispatcher corresponds to no statements executed).
- During the `inlineDispatcherSwitch` pass, in principle we can remove a `case` composed by a single `SwitchBreak` node. However, in the current state of things, this is done before in a dedicated pass called `simplifySwitchBreak`, indeed there is a specific point where we check that we do not encounter any `SwitchBreak` in this phase ([this check here](https://github.com/andrealinux1/revng-c/blob/954765bd27fbbbd5b28e738c30591490f8dc315e/lib/RestructureCFG/InlineDispatcherSwitch.cpp#L707)).
- However, I share your concern that this is not crystal clear to understand.
- At the same time, I think that the semantics of the `containsNonLocalCF` helper function should handle the presence of a `SwitchBreak`. Maybe the non ideal thing here is the name of the helper itself. Basically we want an helper which replies "no you can't move around this `ASTNode`" only when it finds non local control flow related to loops (standard `break`s and `continue`s)

include/revng-c/RestructureCFG/ASTNode.h
andrealinux1: Ohh right, addressed this in 5cc1ef4720c1a3af2055924147c447000280205f

lib/RestructureCFG/InlineDispatcherSwitch.cpp
andrealinux1: Right, changed this [here](https://github.com/revng/revng-c/blob/f15b7b1a9506a37ed6d348f7dd6d880b1d93df16/lib/RestructureCFG/InlineDispatcherSwitch.cpp#L519)

lib/RestructureCFG/InlineDispatcherSwitch.cpp
andrealinux1: There was actually a bug here, the inlining was always enabled, I now disabled it [here](https://github.com/revng/revng-c/blob/f15b7b1a9506a37ed6d348f7dd6d880b1d93df16/lib/RestructureCFG/InlineDispatcherSwitch.cpp#L532), and added a TODO mentioning that we may want to extend the inlining criterion to cover also this scenario.

The main point is that we could consider this advanced scenario at the cost of moving also the eventual exit dispatcher associated to a cycle when moving the cycle itself. If we do that, I think that we may also enable the inlining for `ScsNode`s.

lib/RestructureCFG/InlineDispatcherSwitch.cpp
andrealinux1: Right, applied the suggested changes [here](https://github.com/revng/revng-c/blob/f15b7b1a9506a37ed6d348f7dd6d880b1d93df16/lib/RestructureCFG/InlineDispatcherSwitch.cpp#L539)

lib/RestructureCFG/InlineDispatcherSwitch.cpp
andrealinux1: Implemented the optimization [here](https://github.com/revng/revng-c/blob/f15b7b1a9506a37ed6d348f7dd6d880b1d93df16/lib/RestructureCFG/InlineDispatcherSwitch.cpp#L560)

lib/RestructureCFG/InlineDispatcherSwitch.cpp
andrealinux1: With the introduction of the new `RemoveDeadCode` beautify pass this portion of code want away.

lib/RestructureCFG/InlineDispatcherSwitch.cpp
andrealinux1: I crafted a new commit dedicated to removing all the unnecessary `RootNode` argument b9a8a08f7f6160f57da4506aea7fce6f85550453

lib/RestructureCFG/InlineDispatcherSwitch.cpp
andrealinux1: Done this in a dedicated commit b9a8a08f7f6160f57da4506aea7fce6f85550453

lib/RestructureCFG/InlineDispatcherSwitch.cpp
andrealinux1: In the new version of the code, the parentheses are actually needed [here](https://github.com/revng/revng-c/blob/266c681abb4668354bbbbc5cb6802ad4c583ac30/lib/RestructureCFG/RemoveDeadCode.cpp#L37)

lib/RestructureCFG/InlineDispatcherSwitch.cpp
andrealinux1: Right, I created a new wrapper to hide the internal working details [here](https://github.com/revng/revng-c/blob/4c08f81ad95989ec412c54a03f1e80560eef7a0a/lib/RestructureCFG/InlineDispatcherSwitch.cpp#L226)

lib/RestructureCFG/InlineDispatcherSwitch.cpp
andrealinux1: The `Atomic` meaning was the one you highlighted. Removed it [here](https://github.com/revng/revng-c/blob/4c08f81ad95989ec412c54a03f1e80560eef7a0a/lib/RestructureCFG/InlineDispatcherSwitch.cpp#L870)

lib/RestructureCFG/InlineDispatcherSwitch.cpp
andrealinux1: In rewriting this as the `RemoveDeadCode` beautify pass, I changed the approach here. Basically, I know look for everything that does not `fallthrough`, and this encompasses `noreturn` functions too. Code [here](https://github.com/revng/revng-c/blob/266c681abb4668354bbbbc5cb6802ad4c583ac30/lib/RestructureCFG/RemoveDeadCode.cpp#L36)

lib/RestructureCFG/InlineDispatcherSwitch.cpp
andrealinux1: I implemented a `removeDeadCode` beautify pass in a separate commit 266c681abb4668354bbbbc5cb6802ad4c583ac30, so everything is more autocontained

andrealinux1:
@pfez One additional thing that I want to mention is that with the introduction of the `removeDeadCode` beautify pass, I found, investigated and squashed a couple of additional bugs that were caused by incorrect semantics of the routine that computes the `FallThroughScope` property of each AST node.

You can take a look at these changes in the following commits: 09204508f87acfcf025f818db4c80fbaf9970893, f6bedcf04ea66e64b1b02a4d604c74195eb1f235, 32fdc5b8072648ec0686b2f5017423e5a449cf05.

I think that another couple of eyes are useful on this, since this pass can prune away parts of the AST. I also have a couple of more elaborated observations on this, maybe ping me IRL when reviewing this and we can discuss them.

aleclearmind:
Never call branches `fix/*`. We only have `feature/*` branches.

andrealinux1:
> Never call branches `fix/*`. We only have `feature/*` branches.

@aleclearmind ACK, shall I rename the current branch for this PR?

include/revng-c/RestructureCFG/ASTNode.h
```diff
@@ -405,6 +405,8 @@ class SequenceNode : public ASTNode {
 
   links_container::size_type length() const { return NodeVec.size(); }
 
+  bool isEmpty() const { return NodeVec.empty(); }
```
pfez: If this is really for consistency with STL then `s/isEmpty/empty` please.

lib/RestructureCFG/InlineDispatcherSwitch.cpp
pfez: :+1:

lib/RestructureCFG/InlineDispatcherSwitch.cpp
pfez: > The main point is that we could consider this advanced scenario at the cost of moving also the eventual exit dispatcher associated to a cycle when moving the cycle itself. If we do that, I think that we may also enable the inlining for ScsNodes.

Ok then please expand the comments explaining this. Otherwise, like it is now, it seems that you never want to do it, not that it's a trade-off decision.

lib/RestructureCFG/InlineDispatcherSwitch.cpp
pfez: I understand now, and I agree with your last statement.

Can you propose better naming for the helper?

lib/RestructureCFG/BeautifyGHAST.cpp
```diff
@@ -1024,6 +1025,11 @@ void beautifyAST(const model::Binary &Model, Function &F, ASTTree &CombedAST) {
   RootNode = inlineDispatcherSwitch(CombedAST);
   Dumper.log("after-dispatcher-switch-inlining");
 
+  // Perform the dead code simplification
```
pfez: Please add more explanation about why it makes sense to call this here and not any other places.

lib/RestructureCFG/InlineDispatcherSwitch.cpp
pfez: Well use `!=` and drop the parentheses then

lib/RestructureCFG/InlineDispatcherSwitch.cpp
pfez: Mmmh ok it makes sense like it is. Just note it in the comments.

include/revng-c/RestructureCFG/ASTNode.h
andrealinux1: Changed this [here](https://github.com/revng/revng-c/blob/6af6c8ffa81d1d570010e6e64e260a78eedd4ea9/include/revng-c/RestructureCFG/ASTNode.h#L408)

lib/RestructureCFG/BeautifyGHAST.cpp
andrealinux1: Added some related comments [here](https://github.com/revng/revng-c/blob/6af6c8ffa81d1d570010e6e64e260a78eedd4ea9/lib/RestructureCFG/BeautifyGHAST.cpp#L1028)

lib/RestructureCFG/InlineDispatcherSwitch.cpp
andrealinux1: I motivated this choice, and highlighted that we may want to reverse it, [here](https://github.com/revng/revng-c/blob/6af6c8ffa81d1d570010e6e64e260a78eedd4ea9/lib/RestructureCFG/InlineDispatcherSwitch.cpp#L540)

lib/RestructureCFG/InlineDispatcherSwitch.cpp
andrealinux1: I changed the name of the helper [here](https://github.com/revng/revng-c/blob/6af6c8ffa81d1d570010e6e64e260a78eedd4ea9/lib/RestructureCFG/InlineDispatcherSwitch.cpp#L521), in `containsContinueOrBreak`

lib/RestructureCFG/InlineDispatcherSwitch.cpp
andrealinux1: Right, changed this [here](https://github.com/revng/revng-c/blob/6af6c8ffa81d1d570010e6e64e260a78eedd4ea9/lib/RestructureCFG/RemoveDeadCode.cpp#L37)

lib/RestructureCFG/InlineDispatcherSwitch.cpp
andrealinux1: I added some comments commenting on the motivation for this assertion [here](https://github.com/revng/revng-c/blob/6af6c8ffa81d1d570010e6e64e260a78eedd4ea9/lib/RestructureCFG/RemoveDeadCode.cpp#L49)

lib/RestructureCFG/InlineDispatcherSwitch.cpp
pfez: Much better, thanks!

PR #14, opened 2024-04-29
Disable promotion of two `case`s standard (non-dispatcher) `switch` with no `default` case

andrealinux1:
Fix the promotion logic to not promote standard `switch`es with two `case`s and no `default` present (or the semantics will be corrupted).

### DIFF
--- a/lib/RestructureCFG/SimplifyDualSwitch.cpp
+++ b/lib/RestructureCFG/SimplifyDualSwitch.cpp
@@ -43,9 +43,16 @@ computeSwitchToIfPromotion(SwitchNode *Switch) {
   Case ElectedCase = Invalid;
 
   // We cannot promote switches with more than 2 cases
-  if (Switch->cases_size() > 2) {
+  size_t CaseNum = Switch->cases_size();
+  if (CaseNum > 2)
     return std::nullopt;
-  }
+
+  // We cannot promote non-dispatcher switches with exactly 2 cases and without
+  // default, because it would be semantically incorrect, given that the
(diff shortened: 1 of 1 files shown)


lib/RestructureCFG/SimplifyDualSwitch.cpp
```diff
@@ -42,9 +42,26 @@ computeSwitchToIfPromotion(SwitchNode *Switch) {
...
+    }
+  } else {
+
+    // We cannot promote switches with more than 2 cases, or switches with two
+    // cases that do not have a default case (the promotion would be
+    // semantically incorrect)
+    size_t CasesNum = Switch->cases_size();
+    if (CasesNum > 2 or (CasesNum == 2 and DefaultCase == nullptr)) {
+      return std::nullopt;
+    }
```
pfez: This is correct, strictly speaking. But it's also much more convoluted than it needs to be.

I think the following would be more concise and always correct.

```c++
// We cannot promote switches with more than 2 cases
size_t CaseNum = Switch->cases_size();
if (CaseNum > 2)
  return std::nullopt;

// We cannot promote non-dispatcher switches with exactly 2 cases and without
// default, because it would be semantically incorrect, given that the implicit
// default for a switch without a default is to jump past the switch.
if (Switch->getCondition() != nullptr and CaseNum == 2 and Switch->getDefault() == nullptr)
  return std::nullopt;
```

andrealinux1: Done in 78abb1a4cb266e6932a0462c535feb83756b385e

PR #17, opened 2024-05-07
Fix licence headers

JackCrumpLeys:
The license header on some files refers to a non-existent `LICENSE.md`. Based on context I assume the license that should be linked is the one held at `LICENSE.mit`.


aleclearmind:
Yeah, this is a known issue, I think we already have a commit to fix this somewhere. Thanks for reporting this though!

Keeping this open until we merge it.

aleclearmind:
We'll do this internally, thanks for reporting this though.

PR #1, opened 2024-03-06
Model and LLVM IR to MLIR conversion

vasama-work:

PR #20, opened 2024-05-21
Clift cleanup

vasama-work:
Various Clift cleanup.
* Replaced many uses of `Type` with `ValueType`.
* Renamed `TypeDefinition` to `TypeDefinitionAttr` and made the naming of the derived classes consistent following the pattern `StructTypeAttr`.
* Added the `Clift_` prefix to tablegen definitions where it was missing.
* Changed Clift alias attribute and type alias generation to be more structured. This uses the `$` character as a special marker. E.g. `my_struct$def` for `StructTypeAttr` alias, `my_struct` for the corresponding `DefinedType` and `my_struct$const` for its const-qualified version.
* Reorganised function definitions in `CliftAttributes.cpp` and `CliftTypes.cpp`.

### DIFF
--- a/include/revng-c/mlir/Dialect/Clift/IR/CliftAttributes.td
+++ b/include/revng-c/mlir/Dialect/Clift/IR/CliftAttributes.td
@@ -73,14 +73,14 @@ def EnumFieldAttr : Clift_Attr<"EnumField", "enum_field", [DeclareAttrInterfaceM
   let assemblyFormat = "`<`struct($name, $raw_value)`>`";
 }
 
-def Clift_EnumAttr : Clift_Attr<"Enum", "enum", [DeclareAttrInterfaceMethods<AliasableAttr>, DeclareAttrInterfaceMethods<SizedType, ["getByteSize"]>, TypeDefinition, SubElementAttrInterface]> {
+def Clift_EnumTypeAttr : Clift_Attr<"EnumType", "enum", [DeclareAttrInterfaceMethods<AliasableAttr>, DeclareAttrInterfaceMethods<SizedType, ["getByteSize"]>, TypeDefinitionAttr, SubElementAttrInterface]> {
   let summary = "Clift enum type";
 
   let parameters = (ins "uint64_t":$id, StringRefParameter<>:$name, "mlir::Type":$underlying_type, ArrayRefParameter<"mlir::clift::EnumFieldAttr">:$fields);
 
   let builders = [
     TypeBuilderWithInferredContext<(ins "uint64_t":$ID, "llvm::StringRef":$Name, "mlir::Type":$ElementType, "llvm::ArrayRef<mlir::clift::EnumFieldAttr>":$Fields), [{
-      return EnumAttr::get(ElementType.getContext(), ID, Name, ElementType, Fields);
+      return EnumTypeAttr::get(ElementType.getContext(), ID, Name, ElementType, Fields);
      }]>,
   ];
 
@@ -92,14 +92,14 @@ def Clift_EnumAttr : Clift_Attr<"Enum", "enum", [DeclareAttrInterfaceMethods<Ali
   let assemblyFormat = "`<`struct($id, $name, $underlying_type)  `,` `fields` `=` `[` $fields `]``>` ";
 }
 
-def Clift_TypedefAttr: Clift_Attr<"Typedef", "typedef", [DeclareAttrInterfaceMethods<AliasableAttr>, DeclareAttrInterfaceMethods<SizedType, ["getByteSize"]>, TypeDefinition, SubElementAttrInterface]> {
+def Clift_TypedefTypeAttr: Clift_Attr<"TypedefType", "typedef", [DeclareAttrInterfaceMethods<AliasableAttr>, DeclareAttrInterfaceMethods<SizedType, ["getByteSize"]>, TypeDefinitionAttr, SubElementAttrInterface]> {
   let summary = "Clift typedef type";
 
   let parameters = (ins "uint64_t":$id, StringRefParameter<>:$name, "mlir::clift::ValueType":$underlying_type);
 
   let builders = [
     TypeBuilderWithInferredContext<(ins "uint64_t":$ID, "llvm::StringRef":$Name, "mlir::clift::ValueType":$ElementType), [{
-      return TypedefAttr::get(ElementType.getContext(), ID, Name, ElementType);
+      return TypedefTypeAttr::get(ElementType.getContext(), ID, Name, ElementType);
      }]>,
   ];
 
@@ -111,14 +111,14 @@ def Clift_TypedefAttr: Clift_Attr<"Typedef", "typedef", [DeclareAttrInterfaceMet
   let assemblyFormat = "`<`struct($id, $name, $underlying_type) `>` ";
 }
 
-def Clift_FunctionAttr : Clift_Attr<"Function", "function", [DeclareAttrInterfaceMethods<AliasableAttr>, DeclareAttrInterfaceMethods<SizedType, ["getByteSize"]>, TypeDefinition, SubElementAttrInterface]> {
+def Clift_FunctionTypeAttr : Clift_Attr<"FunctionType", "function", [DeclareAttrInterfaceMethods<AliasableAttr>, DeclareAttrInterfaceMethods<SizedType, ["getByteSize"]>, TypeDefinitionAttr, SubElementAttrInterface]> {
   let summary = "Clift function type";
 
   let parameters = (ins "uint64_t":$id, StringRefParameter<>:$name, "mlir::clift::ValueType":$return_type, OptionalArrayRefParameter<"mlir::clift::FunctionArgumentAttr">:$argument_types);
 
   let builders = [
     TypeBuilderWithInferredContext<(ins "uint64_t":$ID, "mlir::clift::ValueType":$ReturnType,   CArg<"llvm::ArrayRef<mlir::clift::FunctionArgumentAttr>", "{}">:$InputTypes, CArg<"llvm::StringRef", "\"\"">:$Name), [{
-      return FunctionAttr::get(ReturnType.getContext(), ID, Name, ReturnType, InputTypes);
+      return FunctionTypeAttr::get(ReturnType.getContext(), ID, Name, ReturnType, InputTypes);
      }]>,
   ];
 
--- a/include/revng-c/mlir/Dialect/Clift/IR/CliftAttributes.td
+++ b/include/revng-c/mlir/Dialect/Clift/IR/CliftAttributes.td
@@ -24,7 +24,7 @@ def Clift_FunctionArgumentAttr: Clift_Attr<"FunctionArgument", "farg", [DeclareA
   let parameters = (ins "mlir::clift::ValueType":$type, StringRefParameter<>:$name);
 
   let builders = [
-    AttrBuilderWithInferredContext<(ins "mlir::Type":$Type,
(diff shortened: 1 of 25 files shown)


lib/mlir/Dialect/Clift/IR/CliftAttributes.cpp
```diff
@@ -102,11 +141,10 @@ mlir::LogicalResult FieldAttr::verify(EmitErrorType EmitError,
                                       uint64_t Offset,
                                       clift::ValueType ElementType,
                                       llvm::StringRef Name) {
-  if (auto Definition = mlir::dyn_cast<DefinedType>(ElementType)) {
-    if (mlir::isa<FunctionTypeAttr>(Definition.getElementType()))
-      return EmitError() << "Underlying type of field attr cannot be a "
-                            "function type";
-  }
+  if (not isObjectType(ElementType))
+    return EmitError() << "Struct and union field types must be object types.";
```
pfez, line 90: :thinking: this makes it impossible to have struct/union fields that are primitives, enums, pointers, which is wrong.

The only forbidden types should be incomplete types (like you properly check below), and zero-sized types, that you also check below.

lib/mlir/Dialect/Clift/IR/CliftAttributes.cpp
```diff
@@ -246,11 +284,8 @@ FunctionTypeAttr::verify(EmitErrorType EmitError,
                          llvm::StringRef Name,
                          clift::ValueType ReturnType,
                          llvm::ArrayRef<FunctionArgumentAttr> Args) {
-  if (const auto Type = mlir::dyn_cast<DefinedType>(ReturnType)) {
-    if (mlir::isa<FunctionTypeAttr>(Type.getElementType()))
-      return EmitError() << "function type cannot return another function type";
-  }
-
+  if (not isVoid(ReturnType) and not isObjectType(ReturnType))
+    return EmitError() << "Function return type must be an object type or void";
```
pfez: :thinking: this makes it impossible to have return types fields that are primitives, enums, pointers, which is wrong.

The only forbidden types should be incomplete types and function types.

lib/mlir/Dialect/Clift/IR/CliftAttributes.cpp
vasama-work: And function types. Also scalar tuple types actually, which I'm not checking for. I'm borrowing C++ terminology here. Note that *object type* is not *class type* (which includes structs and unions). *Object type* is what you might think of as value type. It's anything that you can have an object of and it includes primitive types, enums, pointers.

lib/mlir/Dialect/Clift/IR/CliftAttributes.cpp
```diff
@@ -23,7 +23,23 @@
...
+[[nodiscard]] static auto onTypeVisited(const uint64_t ID,
+                                        const mlir::Attribute &Attr) {
+  const auto R = CurrentlyVisitedTypes.try_emplace(ID, Attr);
+  revng_assert(R.second);
+
+  return llvm::make_scope_exit([=]() {
+    const auto N = CurrentlyVisitedTypes.erase(ID);
+    revng_assert(N == 1);
+  });
+}
```
aleclearmind: Can you explain what's going on here?
(this belongs to #1)

include/revng-c/mlir/Dialect/Clift/IR/CliftTypes.h
```diff
@@ -15,8 +15,80 @@
...
 #include "revng-c/mlir/Dialect/Clift/IR/CliftOpsTypes.h.inc"
+
+namespace mlir::clift {
+
+struct ScalarTupleTypeStorage;
+
+/// Scalar tuples are used to represent the multiple register return values of
+/// raw function types. They are much like structs, but permit only scalar
+/// element types and have no explicit offsets for their elements.
+class ScalarTupleType : public Type::TypeBase<ScalarTupleType,
```
aleclearmind, line 35: Add a note about the fact that this is temporary, eventually we'll let RFTs return plain structs and get rid of this.
(this belongs to #1)

lib/mlir/Dialect/Clift/IR/CliftTypes.cpp
```diff
@@ -59,21 +59,50 @@ void CliftDialect::printType(mlir::Type Type,
...
+      return PrimitiveKind::VoidKind;
+
+    case model::PrimitiveTypeKind::Invalid:
+    case model::PrimitiveTypeKind::Count:
+      // Unreachable. This causes an error during constant evaluation.
+      return UninitializedKind;
+    }
+  };
+
+  for (int I = 0; I < static_cast<int>(model::PrimitiveTypeKind::Count); ++I) {
```
aleclearmind, line 100: It'd be nice to generate a generator function that yields them all. Not urgent though.

aleclearmind:
@pfez this is ready for me. If you want to perform further review, remember to start from `Remove redundant global namespace qualifiers`, the ones before are from #1.

aleclearmind (changes requested):

include/revng-c/mlir/Dialect/Clift/IR/CliftAttributes.td
```diff
@@ -15,7 +15,7 @@ class Clift_Attr<string name, string attrMnemonic, list<Trait> traits = []>
   let mnemonic = attrMnemonic;
 }
 
-def FunctionArgumentAttr: Clift_Attr<"FunctionArgument", "farg", [DeclareAttrInterfaceMethods<SubElementAttrInterface>]> {
```
pfez, line 18: Brief recap of why this is necessary:
- you tried removing the `Clift_` prefix from everywhere, and doing that doesn't work because the un-prefixed names clash with other auto-generated stuff we can't control
- so we decided to apply the prefix to everything for consistency

One question though: can the prefix be just `Clift` instead of `Clift_`? `Clift_` is ugly, because it's not `CamelCase`, nor `snake_case`. I'd be much happier if we can make it `CamelCase` like everything else.

lib/mlir/Dialect/Clift/IR/CliftTypes.cpp
```diff
@@ -59,21 +59,50 @@ void CliftDialect::printType(mlir::Type Type,
...
-static_assert(Primitive::Number == kindToKind(PrimitiveKind::NumberKind));
+static constexpr model::PrimitiveTypeKind::Values
+kindToKind(const PrimitiveKind Kind) {
+  return static_cast<model::PrimitiveTypeKind::Values>(Kind);
+}
+
+/// Test that kindToKind converts each clift::PrimitiveKind to the matching
+/// model::PrimitiveTypeKind. Use a switch converting in the opposite direction
+/// in order to produce a warning if a new primitive kind is ever added.
+static consteval bool testKindToKind() {
```
pfez, line 74: Hurray! The first use of `consteval` in `revng-c`!

include/revng-c/mlir/Dialect/Clift/IR/CliftTypes.td
```diff
@@ -43,7 +43,8 @@ def Clift_PrimitiveType : Clift_Type<"Primitive", "primitive", [Clift_ValueType,
   let assemblyFormat = "`<`( `is_const` `=` $is_const^ `,`` `)? $kind $size `>` ";
 }
 
-def Clift_PointerType : Clift_Type<"Pointer", "pointer", [DeclareTypeInterfaceMethods<Clift_ValueType, ["getByteSize"]>, SubElementTypeInterface]> {
+
+def Clift_PointerType : Clift_Type<"Pointer", "pointer", [DeclareTypeInterfaceMethods<Clift_ValueType, ["getByteSize"]>, SubElementTypeInterface, DeclareTypeInterfaceMethods<Clift_AliasableType>]> {
```
pfez: I don't think we agreed to make pointer types (and array types later) aliasable.

Last time we talked about this we said we want pointers and arrays to always be emitted inline, so no alias is allowed.

Did you find any reason that make this necessary, or was it just for consistency with the others?

include/revng-c/mlir/Dialect/Clift/IR/CliftTypes.td
```diff
@@ -78,7 +79,7 @@ def Clift_DefinedType : Clift_Type<"Defined", "defined", [DeclareTypeInterfaceMe
   let assemblyFormat = "`<`( `is_const` `=` $is_const^ `,`` ` )? $element_type `>` ";
 }
 
-def Clift_ArrayType : Clift_Type<"Array", "array", [DeclareTypeInterfaceMethods<Clift_ValueType, ["getByteSize"]>, SubElementTypeInterface]> {
+def Clift_ArrayType : Clift_Type<"Array", "array", [DeclareTypeInterfaceMethods<Clift_ValueType, ["getByteSize"]>, SubElementTypeInterface, DeclareTypeInterfaceMethods<Clift_AliasableType>]> {
```
pfez: Here's the same issue for array types. They shouldn't be aliasable.

lib/mlir/Dialect/Clift/IR/CliftTypes.cpp
```diff
@@ -145,6 +147,13 @@ uint64_t PointerType::getByteSize() const {
   return getPointerSize();
 }
 
+bool PointerType::getAlias(llvm::raw_ostream &OS) const {
+  // Aliases for pointers could be generated, but that would currently result in
+  // incorrect ordering of printed aliases when recursive types are involved. To
+  // fix this might require changes to how incomplete types are represented.
+  return false;
+}
+
```
pfez: So this should go away.

lib/mlir/Dialect/Clift/IR/CliftTypes.cpp
```diff
@@ -166,6 +175,14 @@ uint64_t ArrayType::getByteSize() const {
...
 }
 
+bool ArrayType::getAlias(llvm::raw_ostream &OS) const {
+  auto Aliasable = mlir::dyn_cast<AliasableType>(getElementType());
+  if (not Aliasable or not Aliasable.getAlias(OS))
+    return false;
+  OS << "$array" << getElementsCount();
+  return true;
+}
+
```
pfez: And this should go away too.

include/revng-c/mlir/Dialect/Clift/IR/CliftTypes.td
vasama-work: I originally intended to do aliases for pointers too, but ran into issues wrt. recursive types. I removed it now.

include/revng-c/mlir/Dialect/Clift/IR/CliftAttributes.td
vasama-work: I don't know if there is a way to configure this, but by default it removes the `Clift_` prefix from the generated class names, but not just `Clift`.

vasama-work: https://mlir.llvm.org/docs/DefiningDialects/Operations/#class-name-and-namespaces

vasama-work: https://github.com/revng/llvm-project/blob/master/mlir/lib/TableGen/Operator.cpp#L41

pfez: So `Clift_FunctionArgumentAttr` causes TableGen to generate `clift::FunctionArgumentAttr` in c++, right?

If so, it makes sense, and feel free to close this thread.

vasama-work: That's my understanding of it, yes.

PR #21, opened 2024-05-22
Drop sections

aleclearmind:


share/revng/test/configuration/revng-c/decompilation.yml
```diff
@@ -3,19 +3,42 @@
...
+    type: revng-c.decompile-to-single-file.check-decompiled-c
+    from:
+      - type: revng-qa.compiled-with-debug-info
+        filter: for-decompilation
+      - type: revng-c.decompile-to-single-file
+    suffix: /
+    command: |-
+      RESUME=$$(temp -d);
+      cp -Tar "$INPUT2" "$$RESUME";
+      revng artifact --resume "$$RESUME" decompile-to-single-file "$INPUT1" | revng ptml > "$OUTPUT/decompiled.c";
```
pfez, line 30: Why are you doing `revng artifact decompile-to-single-file -o /dev/null` above and then redoing the whole thing here? Wouldn't it be better to save the `.c` file above and avoid one call of `revng artifact` here?

share/revng/test/configuration/revng-c/decompilation.yml
```diff
@@ -3,19 +3,42 @@
...
+  - # Perform checks on the decompiled C code
+    type: revng-c.decompile-to-single-file.filecheck
+    from:
+      - type: revng-qa.compiled-with-debug-info
+        filter: for-decompilation
+      - type: revng-c.decompile-to-single-file
+    command: |-
+      RESUME=$$(temp -d);
+      cp -Tar "$INPUT2" "$$RESUME";
+      revng artifact --resume "$$RESUME" decompile-to-single-file "$INPUT1" | revng ptml | FileCheck "${SOURCE}".filecheck
```
pfez, line 44: And potentially you can save this too.

share/revng/test/configuration/revng-c/decompilation.yml
aleclearmind: This invocation of `revng artifact` should be very fast, given that we already have the result in the `--resume` directory.

The command above is just to "warm up" the `--resume` directory basically.

share/revng/test/configuration/revng-c/decompilation.yml
aleclearmind: Let's discuss in the other comment.

share/revng/test/configuration/revng-c/decompilation.yml
pfez: Yeah, but until we don't switch to the new MLIR-based backend it will redo the CFG restructuring every time, which still takes time. I'd prefer to shave off time in tests to improve QoL of developers.

aleclearmind: No, it doesn't redo anything. Everything is cached in the `--resume` directory. The command will basically fetch things from there and serve them on output.

pfez: Ah nice then!

PR #25, opened 2024-06-12
Clift statements

vasama-work:


lib/mlir/Dialect/Clift/IR/CliftOps.cpp
```diff
@@ -92,83 +96,140 @@ void clift::ModuleOp::build(OpBuilder &Builder, OperationState &State) {
...
+  // Visit a field type of a class type attribute.
+  // RootAttr is the root class type attribute and is used to detect recursion.
+  mlir::LogicalResult visitFieldType(mlir::Operation *ContainingOp,
+                                     clift::ValueType FieldType,
+                                     TypeDefinitionAttr RootAttr) {
+    FieldType = dealias(FieldType);
+
+    if (auto T = mlir::dyn_cast<DefinedType>(FieldType)) {
+      if (T.getElementType() == RootAttr)
+        return ContainingOp->emitError() << "Clift ModuleOp contains a "
```
pfez, line 142: Ok fine now we are back to being able to detect self-loops. But what about larger loops?

What happens if `struct A` has a field with type `struct B`, which in turn has a field with type `struct A`?

I see that there aren't any unit tests on this.

lib/mlir/Dialect/Clift/IR/CliftOps.cpp
```diff
@@ -190,27 +252,27 @@ struct ModuleValidator {
...
+
     return mlir::success();
   }
 
 private:
-  llvm::SmallPtrSet<mlir::Type, 2> VisistedTypes;
-  llvm::SmallPtrSet<mlir::Attribute, 2> VisitedAttrs;
-  llvm::DenseMap<uint64_t, mlir::clift::TypeDefinitionAttr> Definitions;
+  llvm::SmallPtrSet<mlir::Type, 32> VisitedTypes;
+  llvm::SmallPtrSet<mlir::Attribute, 32> VisitedAttrs;
```
pfez, line 341: Why 32 now?

Not that I necessarily dislike that, but is it any less arbitrary than 2 or what?

lib/mlir/Dialect/Clift/IR/CliftOps.cpp
```diff
@@ -25,6 +25,28 @@ void mlir::clift::CliftDialect::registerOperations() {
...
+  });
+
+  return not Result.wasInterrupted();
+}
+
+template bool clift::impl::verifyRegionContent<StatementOpInterface>(Region &R);
+
+template bool
+clift::impl::verifyRegionContent<ExpressionOpInterface>(Region &R);
+
```
pfez: It's very generic defined like this, but I think we could use `verifyRegionContent` to make sure that no operation is both a statement and an expression.

Something like:
```c++
template<bool OnlyStatements>
bool clift::impl::verifyRegionContent(Region &R) {
  if (R.empty())
    return true;

  if (not R.hasOneBlock())
    return false;
  
  using AllowedOpInterface = std::conditional_t<OnlyStatements, StatementOpInterface, ExpressionOpInterface>;
  using ForbiddenOpInterface = std::conditional_t<not OnlyStatements, ExpressionOpInterface, StatementOpInterface;
  const auto Result = R.walk([&](Operation *const NestedOp) -> WalkResult {
    return mlir::success(mlir::isa<AllowedOpInterface>(NestedOp) and not mlir::isa<ForbiddenOpInterface>(NestedOp));
  });

  return not Result.wasInterrupted();
}
```

Then it would be nice to have a way to properly report errors whenever e.g. an operation is both a statement and an expression. Not that I expect anything like that to pass review, but during developement and onboarding of junior devs it can be very helpful.

lib/mlir/Dialect/Clift/IR/CliftOpTraits.cpp
```diff
@@ -0,0 +1,21 @@
...
+using namespace mlir;
+using namespace mlir::clift;
+
+namespace traits_impl = OpTrait::clift::impl;
+
+LogicalResult traits_impl::verifyNoFallthroughTrait(Operation *const Op) {
+  if (Block *const B = Op->getBlock()) {
+    Operation *const NextOp = B->getOperations().getNextNode(*Op);
+    if (NextOp != nullptr and not mlir::isa<AssignLabelOp>(NextOp))
+      return Op->emitOpError() << "Operation may not be succeeded by a "
```
pfez: `s/succeeded/followed`

include/revng-c/mlir/Dialect/Clift/IR/CliftOps.td
```diff
@@ -26,6 +27,120 @@ def Clift_StatementRegion : Region<CPred<"::mlir::clift::impl::verifyRegionConte
...
+def Clift_ModuleOp : Clift_Op<"module",
+                              [SymbolTable,
+                               HasOnlyGraphRegion,
+                               NoRegionArguments,
+                               NoTerminator,
+                               SingleBlock,
+                               RegionKindInterface]> {
+  let regions = (region AnyRegion:$body);
+
+  let skipDefaultBuilders = 1;
```
pfez, line 54: I know this was present even before your changes, but can you please refresh what it does and why we need it?

include/revng-c/mlir/Dialect/Clift/IR/CliftOps.td
```diff
@@ -26,6 +27,120 @@ def Clift_StatementRegion : Region<CPred<"::mlir::clift::impl::verifyRegionConte
...
+    $body attr-dict
+  }];
+
+  let hasVerifier = 1;
+}
+
+def Clift_FunctionOp : Clift_Op<"func",
+                                [Symbol,
+                                 CallableOpInterface,
+                                 FunctionOpInterface,
```
pfez, line 67: I see that you have done the homework on figuring out what of the methods defined below are part of these interfaces. Thanks!

include/revng-c/mlir/Dialect/Clift/IR/CliftOps.td
```diff
@@ -26,6 +27,120 @@ def Clift_StatementRegion : Region<CPred<"::mlir::clift::impl::verifyRegionConte
...
+                                 AutomaticAllocationScope,
+                                 IsolatedFromAbove,
+                                 HasParent<"mlir::clift::ModuleOp">]> {
+  let arguments = (ins StrAttr:$sym_name,
+                       TypeAttrOf<FunctionType>:$function_type,
+                       OptionalAttr<DictArrayAttr>:$arg_attrs,
+                       OptionalAttr<DictArrayAttr>:$res_attrs);
```
pfez: This is different from the spec. Arguments and return types should be entirely implied by the `FunctionType`.

include/revng-c/mlir/Dialect/Clift/IR/CliftOps.td
```diff
@@ -26,6 +27,120 @@ def Clift_StatementRegion : Region<CPred<"::mlir::clift::impl::verifyRegionConte
...
+}
+
+def Clift_FunctionOp : Clift_Op<"func",
+                                [Symbol,
+                                 CallableOpInterface,
+                                 FunctionOpInterface,
+                                 AutomaticAllocationScope,
+                                 IsolatedFromAbove,
+                                 HasParent<"mlir::clift::ModuleOp">]> {
+  let arguments = (ins StrAttr:$sym_name,
```
pfez, line 71: Also, `UniqueHandle` and `Kind` are missing. Let's discuss this tomorrow.

lib/mlir/Dialect/Clift/IR/CliftOps.cpp
```diff
@@ -234,65 +235,76 @@ mlir::LogicalResult clift::ModuleOp::verify() {
...
-mlir::clift::AssignLabelOp::canonicalize(mlir::clift::AssignLabelOp Op,
-                                         mlir::PatternRewriter &Rewriter) {
-  for (const mlir::OpOperand &Use : Op.getLabel().getUses())
-    if (mlir::isa<mlir::clift::GoToOp>(Use.getOwner()))
-      return mlir::success();
+//===----------------------------- MakeLabelOp ----------------------------===//
 
-  Rewriter.eraseOp(Op);
-  return mlir::success();
+static std::pair<size_t, size_t> getMakeLabelOpUsers(MakeLabelOp Op) {
```
pfez: Rename the function to `getNumLabelUsers` or `countLabelUsers` please (`getNum` is ugly but it's typical in LLVM). The current name suggests you're getting users, while you're only counting them.

include/revng-c/mlir/Dialect/Clift/IR/CliftOps.td
```diff
@@ -141,73 +141,125 @@ def Clift_UndefOp: Clift_Op<"undef"> {
...
 
   let assemblyFormat = [{
-    $label type($label) attr-dict
+    $type (`=` $initializer^)? attr-dict
   }];
 }
 
-def Clift_MakeLabelOp : Clift_Op<"make_label", [Clift_OneUseOfType<"AssignLabelOp">] > {
-  let results = (outs Res<Clift_LabelType, "", [MemAlloc, MemWrite] >:$result);
+def Clift_MakeLabelOp : Clift_Op<"make_label",
```
pfez: Can you keep this close to `goto` and `assign_label`? They belong together.

include/revng-c/mlir/Dialect/Clift/IR/CliftOps.td
```diff
@@ -141,73 +141,125 @@ def Clift_UndefOp: Clift_Op<"undef"> {
...
+                                 [Clift_StatementOpInterface,
+                                  Clift_OneUseOfType<"mlir::clift::AssignLabelOp">] > {
+  let results = (outs Clift_LabelType:$result);
 
   let assemblyFormat = [{
     type($result) attr-dict
   }];
+
+  let hasCanonicalizeMethod = 1;
```
pfez: Neat! Moving canonicalization on the label is the right thing to do!

include/revng-c/mlir/Dialect/Clift/IR/CliftOps.td
```diff
@@ -141,73 +141,125 @@ def Clift_UndefOp: Clift_Op<"undef"> {
...
+    [[nodiscard]] Region &getConditionRegion() {
+      return getRegion(0);
+    }
+
+    [[nodiscard]] Region &getDefaultCaseRegion() {
+      return getRegion(1);
+    }
+
+    [[nodiscard]] unsigned getNumCases() {
+      return getNumRegions() - 2;
```
pfez, line 317: We should check what e.g. CIR does on structured switches. It's not obvious if you want to count the default as a case or not.

pfez (changes requested):

lib/mlir/Dialect/Clift/IR/CliftOps.cpp
vasama-work: No, it's completely arbitrary. 2 just seemed completely pointless. I'm not sure if `SmallPtrSet` has a default for it. I could just remove it and use the LLVM default if it does.

lib/mlir/Dialect/Clift/IR/CliftOps.cpp
vasama-work: I think we're going to have to split them up into separate functions anyway because the expression region block should be non-empty if present and have a yield as its last operation.

include/revng-c/mlir/Dialect/Clift/IR/CliftOps.td
vasama-work: See here: https://github.com/revng/llvm-project/blob/master/mlir/include/mlir/IR/FunctionInterfaces.td#L37

include/revng-c/mlir/Dialect/Clift/IR/CliftOps.td
vasama-work: I just ordered them alphabetically, but sure I can do that.

include/revng-c/mlir/Dialect/Clift/IR/CliftOps.td
vasama-work: It disables generation of the default builders for this operation. The builders are functions used by `mlir::OpBuilder` when you do e.g. `builder.create<ModuleOp>(loc)`. We want to define our own builder without arguments to give it a custom body. You can see that [here](https://github.com/vasama-work/revng-c/blob/feature/clift-ops/lib/mlir/Dialect/Clift/IR/CliftOps.cpp#L53). The purpose of it is to ensure that a module created in such a way always has a block in its sole region.

include/revng-c/mlir/Dialect/Clift/IR/CliftOps.td
vasama-work: Looking at how it's used in the interface though, maybe we can actually replace these arguments with functions.

lib/mlir/Dialect/Clift/IR/CliftOps.cpp
vasama-work: Looks like it does not. Also, 32 is the largest value that `SmallPtrSet` permits. I think that was why I chose it.

lib/mlir/Dialect/Clift/IR/CliftOps.cpp
vasama-work: As for checking that no operation implements both interfaces, I think this is absolutely not the place to do it. If you want, I can perhaps add a unit test to iterate over all Clift dialect operations and check that none of them implement both.

vasama-work: I added those expression region checks I mentioned, but not the one you asked for.

pfez:
Commit 3f99a4bae48b78f0c42bafe7a1c4b56ffca029b6 is empty. Did you lose something, or is it just a leftover?

include/revng-c/mlir/Dialect/Clift/IR/CliftOps.td
pfez: Cool thanks!

include/revng-c/mlir/Dialect/Clift/IR/CliftOps.td
pfez: We have already discussed this at lenght. Closing this thread. I'll review the implementation and comment on specific issues if relevant.

lib/mlir/Dialect/Clift/IR/CliftOps.cpp
```diff
@@ -25,6 +25,42 @@ void mlir::clift::CliftDialect::registerOperations() {
                 /* End of operations list */>();
 }
 
+//===---------------------------- Region types ----------------------------===//
+
+template<typename OpInterface>
+static bool verifyRegionContent(Region &R, const bool Required) {
```
pfez, line 63: Seems like `Required` could be a `bool` template parameter.

include/revng-c/mlir/Dialect/Clift/IR/CliftOps.td
pfez: Yes please. Keep things that are logically related close together.

lib/mlir/Dialect/Clift/IR/CliftOps.cpp
```diff
@@ -273,15 +312,48 @@ struct ModuleValidator {
...
+
+    if (mlir::isa<SwitchOp>(Op))
+      return ControlStatementKind::Switch;
+
+    return std::nullopt;
+  }
+
+  bool hasParentStatement(Operation *Op,
+                          ControlStatementKind Kind,
+                          bool DirectlyNested = true) {
```
pfez: Why default to `true`?
I think this is error prone (see my comments on some of its uses).
I'd go for not defaulting this argument at all. So everyone who calls this will have to ask if they want a direct parent or transitive parent.

lib/mlir/Dialect/Clift/IR/CliftOps.cpp
```diff
@@ -261,6 +271,35 @@ struct ModuleValidator {
...
+        return Op->emitOpError()
+               << Op->getName() << " must be nested within a switch operation.";
+    } else if (mlir::isa<LoopBreakOp>(Op)) {
+      if (not hasParentStatement(Op, ControlStatementKind::Loop))
+        return Op->emitOpError()
+               << Op->getName() << " must be nested within a loop operation.";
+    } else if (mlir::isa<LoopContinueOp>(Op)) {
+      if (not hasParentStatement(Op, ControlStatementKind::Loop, false))
+        return Op->emitOpError()
+               << Op->getName() << " must be nested within a loop operation.";
```
pfez: It seems to me that we all the calls to `hasParentStatement` for `LoopBreakOp` and `SwitcBreakOp` are wrong.

Think about:
```c
while (cond) {
  if (x)
    break; // LoopBreakOp
  else
    continue;
  do_stuff();
}
```
and
```c
switch (val) {
  case 0:
    if (cond)
      break; // SwitchBreakOp
    do_stuff();
}
```

In both cases `hasParentStatement` for the `break`s return `false` if you only allow it for direct parents.

And that is wrong, because both are legitimate `break`s.

They should probably be rewritten as:
```c++
    } else if (mlir::isa<SwitchBreakOp>(Op)) {
      if (hasParentStatement(Op, ControlStatementKind::Loop, false)
        or not HasParentStatement(Op, ControlStatementKind::Switch, false))
        return Op->emitOpError()
               << Op->getName() << " must be nested within a switch operation.";
    } else if (mlir::isa<LoopBreakOp>(Op)) {
      if (hasParentStatement(Op, ControlStatementKind::Switch, false)
        or not HasParentStatement(Op, ControlStatementKind::Loop, false))
        return Op->emitOpError()
               << Op->getName() << " must be nested within a loop operation.";
    }
```

pfez: In hindsight, the third parameter is always `false`, perhaps we don't need it at all?

include/revng-c/mlir/Dialect/Clift/IR/CliftOps.td
```diff
@@ -160,6 +160,17 @@ def Clift_DoWhileOp : Clift_Op<"do_while",
   let assemblyFormat = [{
     $body $condition attr-dict
   }];
+
+  let hasVerifier = 1;
+}
+
+def Clift_ExpressionOp : Clift_Op<"expr",
```
pfez: This is the expression statement right?

I'd call it `Clift_ExpressionStatementOp`, to be extra explicit and avoid confusion.

lib/mlir/Dialect/Clift/IR/CliftOps.cpp
```diff
@@ -263,6 +297,34 @@ mlir::LogicalResult ForOp::verify() {
...
+    return emitOpError() << getOperationName()
+                         << " condition requires a scalar type.";
+
+  return mlir::success();
+}
+
+//===--------------------------- LocalVariableOp --------------------------===//
+
+mlir::LogicalResult LocalVariableOp::verify() {
+  if (Region &R = getInitializer(); not R.empty()) {
```
pfez, line 501: Isn't the initializer region optional? Isn't `getInitializer()` going to fail then? (same for `GlobalVariablOp::verify()`)?

include/revng-c/mlir/Dialect/Clift/IR/CliftOps.td
```diff
@@ -227,10 +227,7 @@ def Clift_SwitchOp : Clift_Op<"switch",
                         Clift_StatementRegion:$default,
                         VariadicRegion<Clift_StatementRegion>:$cases);
```
pfez, line 294: I think we should be able to use `SameVariadicOperandSize`

See how ClangIR does [here](https://github.com/llvm/clangir/blob/main/clang/include/clang/CIR/Dialect/IR/CIROps.td#L1830C8-L1830C31)

lib/mlir/Dialect/Clift/IR/CliftOps.cpp
```diff
@@ -307,6 +307,70 @@ void SwitchOp::build(OpBuilder &OdsBuilder,
...
+  if (hasDefaultCase()) {
+    Printer << " default ";
+    Printer.printRegion(getDefaultCaseRegion());
+  }
+
+  static constexpr llvm::StringRef Elided[] = {
+    "case_values",
+  };
+
+  Printer.printOptionalAttrDict(getOperation()->getAttrs(), Elided);
```
pfez, line 637: :thinking: I'm not sure I understand this, and the parser associated part.
I've also looked at the tests and I still don't understand the serialization format: what's the `case_values` part in `switch-with-mismatched-cases.mlir` required for?

lib/mlir/Dialect/Clift/IR/CliftOps.cpp
```diff
@@ -251,6 +251,21 @@ mlir::LogicalResult clift::ModuleOp::verify() {
 
 //===----------------------------- Statements -----------------------------===//
 
+//===-------------------------------- ForOp -------------------------------===//
+
+mlir::LogicalResult ForOp::verify() {
```
pfez, line 469: About `verify()` for loops in general (`while`, `do`-`while`, `for`): I think we should also verify that the `condition` expression region yields an integer or pointer type.

lib/mlir/Dialect/Clift/IR/CliftOps.cpp
```diff
@@ -385,6 +392,46 @@ mlir::LogicalResult clift::ModuleOp::verify() {
...
+  return {};
+}
+
+mlir::Attribute FunctionOp::removeResAttrsAttr() {
+  return {};
+}
+
+Type FunctionOp::cloneTypeWith(TypeRange inputs, TypeRange results) {
+  revng_abort();
+}
```
pfez: Please, add an `"Operation not supported"` error message here too.

include/revng-c/mlir/Dialect/Clift/IR/CliftOps.td
pfez: Ok this is done now.

lib/mlir/Dialect/Clift/IR/CliftOps.cpp
pfez: Yeah we've discussed about this in person. It's fine like this.

lib/mlir/Dialect/Clift/IR/CliftOps.cpp
vasama-work: `hasParentStatement` only considers loops and switches so nesting within an if-statement has no effect. I added a comment explaining this.

lib/mlir/Dialect/Clift/IR/CliftOps.cpp
vasama-work: The region always exists. The block might not.

lib/mlir/Dialect/Clift/IR/CliftOps.cpp
vasama-work: `switch-with-mismatched-cases.mlir` uses the universal MLIR format to specify `case_values` explicitly. That's required for testing mismatched case values and regions, because using the custom assembly format it is impossible to define such an operation.

lib/mlir/Dialect/Clift/IR/CliftOps.cpp
vasama-work: We do. It's added in the subsequent commit that also adds the minimal expression ops, because without the expressions there is nothing to test. See [here](https://github.com/revng/revng-c/pull/25/commits/c107f733663ad7a0475c8632fca552af66333639#diff-6b10b0ccf89a76f6de01b3fb1d1c780894a2d8a299beba9ac22cf98471b05383R300).

lib/mlir/Dialect/Clift/IR/CliftOps.cpp
vasama-work: I defaulted it to `true` because a) that's the more conservative option, i.e. more likely to not find the required parent, and b) it's used more often.

Still, I've removed the default argument.

include/revng-c/mlir/Dialect/Clift/IR/CliftOps.td
vasama-work: From reading the LLVM source, I can't even find an implementation of this, but further it seems to only look at *operands*, while the array of case values is an *attribute*. That's my understanding of it anyway.

lib/mlir/Dialect/Clift/IR/CliftOps.cpp
vasama-work: It's a regular parameter of `verifyExpressionRegion`. I see no reason to jump through hoops to make it a template parameter of `verifyRegionContent`.

lib/mlir/Dialect/Clift/IR/CliftOps.cpp
pfez: Perfect!

lib/mlir/Dialect/Clift/IR/CliftOps.cpp
pfez: Eh ok. But then `hasParentStatement` does not do what it advertises on the surface. Adding the comment definitely helps. But there's another issue that I didn't really brought up but that is important: I know we're going to need something like `hasParentStatement` in the future for working not just on `switch`es and loops. And `hasParentStatement` would be a perfect name for that, and it also shares 99% of the implementation. It's just that at the moment we're kind-of-artificially restricting it to only be able to reason about `switch`es and loops.

I think that we should have a `hasParentStatement` should be capable of handling any statement, and have 2 modes of operations: one-step (only checks direct parent, return `false` if the direct parent does not exist), transitive (checks all parents, to se if finds at some point one of the requested kind.

Then, the specific checks you need to perform for verifying the well-nestedness of `break`s inside loops of `switch`es can be expressed using the new `hasParentStatement`.

lib/mlir/Dialect/Clift/IR/CliftOps.cpp
pfez: Thanks.

lib/mlir/Dialect/Clift/IR/CliftOps.cpp
pfez: What's the universal MLIR format

Even without knowing what that is, we're implementing custom serialization/deserialization for `switch`, so I don't think we have any obligation to follow its rules.

Honestly, I'd be much happier if our parser didn't recognize a deserialization format that a) is ugly and unintuitive to read and b) allows parsing stuff that can not verify.
I know that there are probably other things that we can parse and do not verify, but it seems to me that disallowing this format would still allow to express with the other text format (the readable one) any valid `switch` construct that we want to allow.
If so, let's just disable the universal format if possible.

And for what concern the tests, I know that without the universal MLIR format we wouldn't be able to test mismatched cases in `.mlir` files anymore, but that just means that the proper way to test this is still C++ and not `.mlir` files + `FileCheck`.

lib/mlir/Dialect/Clift/IR/CliftOps.cpp
pfez: This is really not a big deal. But I'm trying to understand what are the limits of what we can do with MLIR, TableGen and ODS.

You say that `Required` is a regular parameter of `verifyExpressionRegion` and that's true.
But my understanding is that it's only so because of the following `TableGen` lines in `CliftOps.td`:
```
def Clift_ExpressionRegion : Region<CPred<"::mlir::clift::impl::verifyExpressionRegion($_self, true)">,
                                    "Region representing an expression">;

def Clift_OptionalExpressionRegion : Region<CPred<"::mlir::clift::impl::verifyExpressionRegion($_self, false)">,
                                            "Optional region representing an expression">;
```

Could we change those line to the following? (now with a template argument for `verifyExpressionRegion`)
```
def Clift_ExpressionRegion : Region<CPred<"::mlir::clift::impl::verifyExpressionRegion<true>($_self)">,
                                    "Region representing an expression">;

def Clift_OptionalExpressionRegion : Region<CPred<"::mlir::clift::impl::verifyExpressionRegion<false>($_self)">,
                                            "Optional region representing an expression">;
```

If so, I would prefer it.

vasama-work: We could, but it requires either moving the functions into the header entirely, or explicitly declaring and instantiating the template for both true and false. I can do that if you prefer, but like I said, it's more hoops to jump through. I personally don't see any benefit in making it a template parameter.

lib/mlir/Dialect/Clift/IR/CliftOps.cpp
vasama-work: > What's the universal MLIR format

The language reference describes the MLIR notation. See https://mlir.llvm.org/docs/LangRef/#operations `generic-operation`. It's generic notation that can describe any operation, regardless of any custom assembly format.

> Honestly, I'd be much happier if our parser didn't recognize [the generic format]

I'm not sure there's anything we can do to disable it, short of modifying LLVM to remove it entirely. Neither am I convinced that it would be a good idea, even if we could easily do it. This is very much an escape hatch where the user is explicitly stepping outside of the guardrails.

> that just means that the proper way to test this is still C++

I don't agree. In this case the argument about pinning the custom assembly format doesn't even apply. Writing the same test in C++ is going to make it much more verbose and more difficult to maintain. To test the thing properly, even the C++ test should check that the verification fails *for the right reason*, which means checking the generated error message, so we're still pinning the error message. Not that I consider this a problem anyway. It's a very simple thing to change the error string embedded in the test if it's changed in the verification.

lib/mlir/Dialect/Clift/IR/CliftOps.cpp
pfez: No big deal, let's leave it like this.

Thanks for the explanation.

lib/mlir/Dialect/Clift/IR/CliftOps.cpp
pfez: For what concerns tests and pinning the error messages: I'm not saying that doing the test in C++ would set us free from the bounds of pinning error messages. And to be clear I think it's fine to pin error messages in this case.

What I was trying to say is:
1. I consider a superior goal to only have a single valid syntax for parsing, even more so if that syntax makes it impossible by design to write `switch` constructs with mismatched case values.
2. If we can get rid of the universal textual format for operations that we define custom parsers for, then that would make it impossible to test in `.mlir` files for `switch` operations with mismatched case values, because they wouldn't be expressible with a valid syntax
3. Even if we succeed in banning the universal parsing format for operations with custom serialization/deserialization, I still consider important to test for `switch` operations with mismatched cases, because those could be constructed in memory and not for parsing. For those, I think it's better to make the test in C++, because doing it in `.mlir` would force us to enable universal format parsing, which I'd like to avoid. I didn't say that doing the test in C++ was less verbose, or less difficult to maintain, or that it allowed not to pin error messages. I just meant that I consider writing the test in C++ (even if many others can be done in `.mlir` files) an acceptable price to pay if that's the only way to ban the universal parsing format, because the universal parsing format AFAIU isn't doing anything good for us that the custom parsing format can't already do.

---

Meanwhile, I've read some more about the universal MLIR format, form the link you pasted, and other material.
At the moment I really don't understand if there is any way to disable it, but I'd really like you to give a shot at removing it, only for `switch`es, or for operations with custom parsing.

> This is very much an escape hatch where the user is explicitly stepping outside of the guardrails.

I understand, but given that stepping outside the guardrails in this case doesn't do anything good (meaning that everything good you can already express with the custom parsing format), I'd really like users to fail to specify weird `switch`es with this format, and only use the allowed one, which already allows all the expressive power needed in order to express any valid constructs.

vasama-work: I really think this is misguided to even try. There's even an `mlir-opt` flag to print all operations using the generic notation. If anything, the custom assembly format should probably be thought of as a secondary one for pretty printing.

> I'd really like you to give a shot at removing it

How hacky a solution are you going to be okay with? Let's say we can do it, at least in debug mode, by inspecting the call stack in some SwitchOp builder. Is that going too far?

pfez: Ok I've investigated some more on this and I'm backing off on disabling the generic notation.

If there's a flag to force printing all operations with the generic notation, that's also a big red flag to forcing its removal. Basically every piece of the MLIR infrastructure expects that to exist and be a first-class citizen.

pfez: At this point I have few questions on parsing and printing though.

So for `switch`es we've implemented custom serialization and deserialization. 

For printing:
1.How do we make sure we abide when the flag for forcing generic text format is passed? Is it just that the main MLIR printing logic select the "builtin" printer for the generic notation and not our custom print format?
2. If the printing of `case_values` for the generic notation is taken care of by MLIR itself, what does the last statement of `SwitchOp::print` (the one I started this thread from) do?

For parsing, I'll open another comment on the relevant lines.

vasama-work: 1. AFAIK, we don't need to do anything for the generic format, since it is generic. The MLIR library is just going to do its own thing and never involve our code.
2. That line prints an optional attribute dictionary containing all the attributes that were not printed manually as part of the custom assembly format, meaning all attributes except `case_values`. Remember that it is possible to attach arbitrary named attributes to operations.

lib/mlir/Dialect/Clift/IR/CliftOps.cpp
```diff
@@ -183,27 +265,419 @@ struct ModuleValidator {
...
+  }
+
+  Result.attributes.set("case_values",
+                        DenseI64ArrayAttr::get(Parser.getContext(),
+                                               CaseValues));
+
+  if (Parser.parseOptionalAttrDict(Result.attributes).failed())
     return mlir::failure();
+
+  return mlir::success();
```
pfez, line 616: For parsing, let me see if I understand what's happening in these last few lines.

`Result.attributes.set("case_values", ...)` is just registering in the `Result` the `CaseValues` that it has parsed above in the `while` loop.

An then what's the catch with `Parser.ParseOptionalAttrDict`?
My understanding is that we don't allow any optional attribute on the `clift::SwitchOp`, so I expect that to always fail.
Or maybe it only succeeds when there's nothing left to parse, which means this is a way to detect when someone is passing additional unexpected attributes. If this is the case then I'm happy and feel free to close this thread.

include/revng-c/mlir/Dialect/Clift/IR/CliftOps.td
pfez: It's odd.

On the one hand, CIR's `SwitchOp` only has a single range of variadic operands, and also has an optional array attribute to represent cases. So that gave me the impression that using the `SameVariadicOperandSize` like they did could also be useful for us.

On the other hand, looking at the doc and at other uses, what you say is correct. Which means we can't use that. Thank you anyways.

lib/mlir/Dialect/Clift/IR/CliftOps.cpp
pfez: Oh ok, so that little dance with `Elided` is to print everything except `case_values`, because those are already printed beforehand.

Thanks a lot.

lib/mlir/Dialect/Clift/IR/CliftOps.cpp
vasama-work: You can always add arbitrary named attributes to operations. They don't all have to be declared in ODS. Users can use these attributes to attach their own data to operations. One example of how *we* are currently using this is attaching model metadata to LLVMIR function operations under the name `revng.function.entry`.

pfez: Oh ok I tend to forget this.

Thanks

lib/mlir/Dialect/Clift/IR/CliftOps.cpp
vasama: > I know we're going to need something like `hasParentStatement` in the future

I would rather deal with that when those concrete use cases actually materialise and the requirements for such an interface become clearer. Since this function is very local to the module validator, would you be okay with just renaming the functions and enum along the lines of `hasLoopOrSwitchParent`?

vasama-work: > I know we're going to need something like hasParentStatement in the future

I would rather deal with that when those concrete use cases actually materialise and the requirements for such an interface become clearer. Since this function is very local to the module validator, would you be okay with just renaming the functions and enum along the lines of `hasLoopOrSwitchParent`?

pfez: Ok that's fine.

PR #28, opened 2024-07-21
Implement Generic Region Identification Analysis

andrealinux1:


include/revng-c/RestructureCFG/GenericRegion.h
```diff
@@ -0,0 +1,91 @@
...
+
+  using child_container = llvm::SmallVector<GenericRegion *>;
+  using child_iterator = child_container::iterator;
+  using child_range = llvm::iterator_range<child_iterator>;
+
+  using edge_descriptor = tuple_cat_t<revng::detail::EdgeDescriptor<NodeT>,
+                                      make_tuple_t<size_t>>;
+  using edge_descriptor_container = llvm::SmallSetVector<NodeT, 4>;
+  using edge_descriptor_iterator = edge_descriptor_container::iterator;
+  using edge_descriptor_range = llvm::iterator_range<edge_descriptor_iterator>;
```
pfez: `edge_descriptor_*` is very verbose. Can we have just `edge_*` instead?

include/revng-c/RestructureCFG/GenericRegion.h
```diff
@@ -0,0 +1,91 @@
...
+
+#include "revng/Support/Assert.h"
+#include "revng/Support/GraphAlgorithms.h"
+
+#include "revng-c/ADT/STLExtras.h"
+
+template<class NodeT>
+class GenericRegion {
+public:
+  using block_container = llvm::SmallSetVector<NodeT, 4>;
```
pfez, line 24: Having `SetVector` is important because we need both:
- uniqueness
- iteration in the order of insertion

Please add a comment to stress this.

include/revng-c/RestructureCFG/GenericRegion.h
```diff
@@ -0,0 +1,91 @@
...
+public:
+  GenericRegion() {}
+
+  void insertBlock(NodeT Block) { Blocks.insert(Block); }
+
+  bool containsBlock(NodeT Block) { return Blocks.contains(Block); }
+
+  block_iterator block_begin() { return Blocks.begin(); }
+  block_iterator block_end() { return Blocks.end(); }
+  block_range blocks() { return llvm::make_range(block_begin(), block_end()); }
```
pfez: I think that we will also need the `const` qualified equivalents of these methods, because after its calculation most uses of `GenericRegion` will be `const`-qualified, given that most clients are read-only.

include/revng-c/RestructureCFG/GenericRegion.h
```diff
@@ -0,0 +1,91 @@
...
+template<class NodeT>
+class GenericRegion {
+public:
+  using block_container = llvm::SmallSetVector<NodeT, 4>;
+  using block_iterator = block_container::iterator;
+  using block_range = llvm::iterator_range<block_iterator>;
+
+  using child_container = llvm::SmallVector<GenericRegion *>;
+  using child_iterator = child_container::iterator;
+  using child_range = llvm::iterator_range<child_iterator>;
```
pfez: I think a `SmallSetVector` would be better here too.
With just a vector we lose uniqueness. And with just a set we lose deterministic ordering of the children range, which is bad for determinism.

include/revng-c/RestructureCFG/GenericRegionInfo.h
```diff
@@ -0,0 +1,80 @@
...
+
+public:
+  using region_container = llvm::SmallVector<
+    std::unique_ptr<GenericRegion<NodeT>>>;
+  using region_iterator = region_container::iterator;
+  using region_range = llvm::iterator_range<region_iterator>;
+
+  using toplevel_region_container = llvm::SmallVector<GenericRegion<NodeT> *>;
+  using toplevel_region_iterator = toplevel_region_container::iterator;
+  using toplevel_region_range = llvm::iterator_range<toplevel_region_iterator>;
```
pfez: Drop these. We don't need a separate container for top level regions. Top level regions are just regions whose parent `nullptr`.

include/revng-c/RestructureCFG/GenericRegionInfo.h
```diff
@@ -0,0 +1,80 @@
...
+
+private:
+  region_container Regions;
+  toplevel_region_container TopLevelRegions;
```
pfez: Drop this too.

include/revng-c/RestructureCFG/GenericRegionInfo.h
```diff
@@ -0,0 +1,80 @@
...
+
+  void compute(GraphT F);
+
+  region_range regions() {
+    return llvm::make_range(Regions.begin(), Regions.end());
+  }
+
+  toplevel_region_range toplevel_regions() {
+    return llvm::make_range(TopLevelRegions.begin(), TopLevelRegions.end());
+  }
```
pfez: And this can become like `make_filter_range(dereferenceRange(regions()), [](const GenericRegion<NodeT> &R) { return R.getParent() != nullptr; })`.

Or you can even make a method `GenericRegion::isRoot` to make this even more concise.

lib/RestructureCFG/GenericRegionInfo.cpp
```diff
@@ -0,0 +1,200 @@
...
+    }
+  }
+
+  // Populate the children regions. We need to perform this operation in a
+  // separate step in order to have already all the created regions in the step
+  // above
+  for (const auto *TLC : CI.toplevel_cycles()) {
+
+    // For each top level cycle, insert a new top level generic region
+    TopLevelRegions.push_back(CycleToRegionMap.at(TLC));
```
pfez: This will go away once you drop `TopLevelRegions` but it's not needed anymore.

lib/RestructureCFG/GenericRegionInfo.cpp
```diff
@@ -0,0 +1,200 @@
...
+
+template<class GraphT, class GT>
+std::string GenericRegionInfo<GraphT, GT>::print() {
+  std::string Output;
+
+  // Print each top level `GenericRegion`, and for each one explore it using a
+  // DFS strategy
+  size_t RegionIndex = 0;
+  Output += "\nGeneric Region Info Results:\n";
+  for (auto &TopLevelRegion : TopLevelRegions) {
```
pfez: And this will just use `top_level_regions()` instead of `TopLevelRegions`.

include/revng-c/RestructureCFG/GenericRegionInfo.h
```diff
@@ -0,0 +1,80 @@
...
+
+private:
+  void initializeRegions(GraphT F);
+  void electHead(GraphT F);
+  void electRetreatings();
+  std::string print();
```
pfez: Can't this method be `const`?

include/revng-c/RestructureCFG/GenericRegionPass.h
```diff
@@ -0,0 +1,23 @@
...
+  static char ID;
+
+public:
+  GenericRegionPass() : llvm::FunctionPass(ID) {}
+
+  bool runOnFunction(llvm::Function &F) override;
+
+  void getAnalysisUsage(llvm::AnalysisUsage &AU) const override;
+
+  void getResult();
```
pfez: This is a just a stub. How are clients supposed to get the analysis results if this returns `void`?

Please implement this.

lib/RestructureCFG/GenericRegionPass.cpp
```diff
@@ -0,0 +1,33 @@
...
+
+void GenericRegionPass::getAnalysisUsage(llvm::AnalysisUsage &AU) const {
+
+  // This is a read only analysis, that does not touch the IR
+  AU.setPreservesAll();
+}
+
+void GenericRegionPass::getResult() {
+  return;
+}
```
pfez: Implement this properly.

lib/RestructureCFG/GenericRegionPass.cpp
```diff
@@ -0,0 +1,33 @@
...
+static constexpr const char *Flag = "generic-region-info";
+using Reg = llvm::RegisterPass<GenericRegionPass>;
+static Reg X(Flag, "Perform the generic region identification analysis");
+
+bool GenericRegionPass::runOnFunction(llvm::Function &F) {
+
+  // Run the `GenericRegionInfo`
+  GenericRegionInfo<llvm::Function *> RIA;
+  RIA.compute(&F);
+
```
pfez: This should store the result somewhere meaningful, so that client passes that depend on this pass can read it.

lib/RestructureCFG/GenericRegionInfo.cpp
```diff
@@ -0,0 +1,200 @@
...
+template<class GraphT, class GT>
+void GenericRegionInfo<GraphT, GT>::initializeRegions(GraphT F) {
+  // We instantiate the `GenericCycle` analysis and wrap the results in
+  // the region objects
+  CycleInfo CI;
+  CI.compute(*F);
+
+  using CycleT = CycleInfo::CycleT;
+  using Region = GenericRegion<NodeT>;
+  std::map<const CycleT *, Region *> CycleToRegionMap;
```
pfez: You're never iterating on this, just inserting and performing lookups. Make it a `SmallMap`

lib/RestructureCFG/GenericRegionInfo.cpp
```diff
@@ -0,0 +1,200 @@
...
+      }
+    }
+  }
+}
+
+template<class GraphT, class GT>
+void GenericRegionInfo<GraphT, GT>::electHead(GraphT F) {
+
+  // For each `Region`, we perform the election of the head node
+  // 1) Compute the reverse post orde
```
pfez: Typo. Missing a trailing `r`.

lib/RestructureCFG/GenericRegionInfo.cpp
```diff
@@ -0,0 +1,200 @@
...
+  llvm::SmallVector<NodeT> RPOT;
+  llvm::copy(llvm::post_order(F), std::back_inserter(RPOT));
+
+  // 2) Compute the distance of each node from the entry node
+  std::map<NodeT, size_t> ShortestPathFromEntry;
```
pfez: You never iterate on this either. Make it a `SmallMap`.

pfez: Also please clarify that "from entry" here always means from the entry block of the function, not the region.

lib/RestructureCFG/GenericRegionInfo.cpp
```diff
@@ -0,0 +1,200 @@
...
+    // the entry with the minimal shortest path from entry. If it is still a
+    // tie, i.e., there are 2 or more candidate heads with, also, the same
+    // minimal shortest path from entry, then we disambiguate by picking the
+    // head that comes first in RPOT.
+    NodeT Head = HeadCandidates.begin()->first;
+    {
+      size_t MaxNHead = HeadCandidates.begin()->second;
+      auto ShortestPathIt = ShortestPathFromEntry.find(Head);
+      revng_assert(ShortestPathIt != ShortestPathFromEntry.end());
+      size_t ShortestPath = ShortestPathFromEntry.at(Head);
```
pfez: Drop these. In the case you need to look at shortest paths you are going to make a lookup in the map nevertheless. All these running variables are just more mental load for the reader.

lib/RestructureCFG/GenericRegionInfo.cpp
```diff
@@ -0,0 +1,200 @@
...
+      auto HeadEnd = HeadCandidates.end();
+      for (NodeT Block : RPOT) {
+        auto HeadIt = HeadCandidates.find(Block);
+        if (HeadIt != HeadEnd) {
+          const auto &[HeadCandidate, NumIncoming] = *HeadIt;
+          if (NumIncoming > MaxNHead) {
+            Head = HeadCandidate;
+            ShortestPath = ShortestPathFromEntry.at(HeadCandidate);
```
pfez: Then you can also drop this.

lib/RestructureCFG/GenericRegionInfo.cpp
```diff
@@ -0,0 +1,200 @@
...
+          } else if (NumIncoming == MaxNHead) {
+            size_t SP = ShortestPathFromEntry.at(HeadCandidate);
+            if (SP < ShortestPath) {
+              Head = HeadCandidate;
+              ShortestPath = SP;
+            }
```
pfez: And rewrite this as:
```c++
size_t CurrentShortest = ShortestPathFromEntry.at(Head);
size_t CandidateShortest = ShortestPathFromEntry.at(CandidateHead);
if (CandidateShortest < CurrentShortest)
  Head = CandidateHead;
```

lib/RestructureCFG/GenericRegionInfo.cpp
```diff
@@ -0,0 +1,200 @@
...
+    // Obtain the insertion iterator for the `Depth` block element
+    auto ShortestIt = ShortestPathFromEntry.insert({ Block, Depth });
+
+    // If we already had in the map an entry for the current block, we need to
+    // assert that the previously found value for the `Depth` is less or equal
+    // of the `Depth` we are inserting
+    if (ShortestIt.second == false) {
+      revng_assert(ShortestIt.first->second <= Depth);
+    }
+  }
```
pfez, line 93: Shortest path from entry is intended to be a tie breaker for electing a head, but you always compute it, even if it might not be needed.

Three actions points:
1. outline the computation of shortest path map to a standalone function.
2. make `ShortestPathFromEntry` a `std::optional` in this method, and initialize it to `nullopt`.
3. only initialize the map the first time you're sure you really need it for breaking a tie.

lib/RestructureCFG/GenericRegionInfo.cpp
pfez: With the precaution I suggested above on `std::optional` you also need:
```c++
if (not ShortestPathFromEntry.has_value())
  ShortestPathFromEntry = computeShortestPaths(F);
```

And all `ShortestPathFromEntry.` become `ShortestPathFromEntry->`.

pfez (changes requested):

include/revng-c/RestructureCFG/GenericRegion.h
andrealinux1: renamed all these `using` removing the `_descriptor` prefix here https://github.com/revng/revng-c/blob/f69bcce2af89d833714a56a52ae34f3581bae8f2/include/revng-c/RestructureCFG/GenericRegion.h#L37

include/revng-c/RestructureCFG/GenericRegion.h
andrealinux1: I added a comment stressing the properties you highlighted above https://github.com/revng/revng-c/blob/f69bcce2af89d833714a56a52ae34f3581bae8f2/include/revng-c/RestructureCFG/GenericRegion.h#L20

include/revng-c/RestructureCFG/GenericRegion.h
andrealinux1: I added the `const` qualified versions of the methods here https://github.com/revng/revng-c/blob/f69bcce2af89d833714a56a52ae34f3581bae8f2/include/revng-c/RestructureCFG/GenericRegion.h#L63

https://github.com/revng/revng-c/blob/f69bcce2af89d833714a56a52ae34f3581bae8f2/include/revng-c/RestructureCFG/GenericRegion.h#L66

https://github.com/revng/revng-c/blob/f69bcce2af89d833714a56a52ae34f3581bae8f2/include/revng-c/RestructureCFG/GenericRegion.h#L69

include/revng-c/RestructureCFG/GenericRegion.h
andrealinux1: I changed this to a `SmallSetVector` and added a comment justifying the change https://github.com/revng/revng-c/blob/f69bcce2af89d833714a56a52ae34f3581bae8f2/include/revng-c/RestructureCFG/GenericRegion.h#L30

include/revng-c/RestructureCFG/GenericRegionInfo.h
andrealinux1: I went with the solution you suggested below

include/revng-c/RestructureCFG/GenericRegionInfo.h
andrealinux1: This has been dropped as you suggested

include/revng-c/RestructureCFG/GenericRegionInfo.h
andrealinux1: `top_level_regions` now returns the filtered range built as you suggested https://github.com/revng/revng-c/blob/f69bcce2af89d833714a56a52ae34f3581bae8f2/include/revng-c/RestructureCFG/GenericRegionInfo.h#L51

lib/RestructureCFG/GenericRegionInfo.cpp
andrealinux1: Dropped this

lib/RestructureCFG/GenericRegionInfo.cpp
andrealinux1: Changed this here https://github.com/revng/revng-c/blob/f69bcce2af89d833714a56a52ae34f3581bae8f2/lib/RestructureCFG/GenericRegionInfo.cpp#L207

include/revng-c/RestructureCFG/GenericRegionInfo.h
andrealinux1: You are right, this is now `const` https://github.com/revng/revng-c/blob/f69bcce2af89d833714a56a52ae34f3581bae8f2/include/revng-c/RestructureCFG/GenericRegionInfo.h#L59

lib/RestructureCFG/GenericRegionInfo.cpp
andrealinux1: Mmmm, maybe you meant `llvm::SmallDenseMap`? I assumed so, changed this here https://github.com/revng/revng-c/blob/f69bcce2af89d833714a56a52ae34f3581bae8f2/lib/RestructureCFG/GenericRegionInfo.cpp#L38

Also, since `llvm::SmallDenseMap` does not have the useful `at`, I added a simple standalone function to mimic the behavior here https://github.com/revng/revng-c/blob/f69bcce2af89d833714a56a52ae34f3581bae8f2/lib/RestructureCFG/GenericRegionInfo.cpp#L22

lib/RestructureCFG/GenericRegionInfo.cpp
andrealinux1: Fixed here https://github.com/revng/revng-c/blob/f69bcce2af89d833714a56a52ae34f3581bae8f2/lib/RestructureCFG/GenericRegionInfo.cpp#L102

lib/RestructureCFG/GenericRegionInfo.cpp
andrealinux1: I specified the `entry block` thing in the comment https://github.com/revng/revng-c/blob/f69bcce2af89d833714a56a52ae34f3581bae8f2/lib/RestructureCFG/GenericRegionInfo.cpp#L106, and made this a `llvm::SmallDenseMap` https://github.com/revng/revng-c/blob/f69bcce2af89d833714a56a52ae34f3581bae8f2/lib/RestructureCFG/GenericRegionInfo.cpp#L108

lib/RestructureCFG/GenericRegionInfo.cpp
andrealinux1: Yeah, dropped these

lib/RestructureCFG/GenericRegionInfo.cpp
andrealinux1: Dropped this

lib/RestructureCFG/GenericRegionInfo.cpp
andrealinux1: I implemented your suggestion of using a `std::optional` variable and computing it only when needed https://github.com/revng/revng-c/blob/f69bcce2af89d833714a56a52ae34f3581bae8f2/lib/RestructureCFG/GenericRegionInfo.cpp#L108

lib/RestructureCFG/GenericRegionInfo.cpp
andrealinux1: I took your suggestions and delayed the computation of `ShortestPathFromEntry` until really needed https://github.com/revng/revng-c/blob/f69bcce2af89d833714a56a52ae34f3581bae8f2/lib/RestructureCFG/GenericRegionInfo.cpp#L147

lib/RestructureCFG/GenericRegionPass.cpp
andrealinux1: Now the `GenericRegionInfo` object is stored inside the pass and a reference to it returned it by the `getResult` method 

https://github.com/revng/revng-c/blob/2874031652f8d9c4240ed2796912c23c75054aca/include/revng-c/RestructureCFG/GenericRegionPass.h#L14

https://github.com/revng/revng-c/blob/2874031652f8d9c4240ed2796912c23c75054aca/lib/RestructureCFG/GenericRegionPass.cpp#L32

lib/RestructureCFG/GenericRegionPass.cpp
andrealinux1: Fixed here 

https://github.com/revng/revng-c/blob/2874031652f8d9c4240ed2796912c23c75054aca/lib/RestructureCFG/GenericRegionPass.cpp#L32

include/revng-c/RestructureCFG/GenericRegionPass.h
andrealinux1: Yeah, this slipped away, since I'm running the unit tests with `FileCheck` and not consuming the results, I did not see this, fixed this here https://github.com/revng/revng-c/blob/2874031652f8d9c4240ed2796912c23c75054aca/include/revng-c/RestructureCFG/GenericRegionPass.h#L26

lib/RestructureCFG/GenericRegionInfo.cpp
pfez: `SmallDenseMap` even better. Thanks!

lib/RestructureCFG/GenericRegionInfo.cpp
pfez: Great!

lib/RestructureCFG/GenericRegionInfo.cpp
pfez: Neat!

lib/RestructureCFG/GenericRegionInfo.cpp
pfez: Yeah, most of your maps were mapping pointers to pointers. `SmallDenseMap` FTW. Thanks

lib/RestructureCFG/GenericRegionPass.cpp
pfez: I'd like it better to return a `const &`, to prevent passes depending on this to mutate it.

All the rest is perfect.

I'll close the other related comments, and leave just this one open, until this is fixed.

andrealinux1: Right, fixed this here

https://github.com/revng/revng-c/blob/e211fe822643ea506c2fccf69b773fb21cf9065c/include/revng-c/RestructureCFG/GenericRegionPass.h#L26

pfez: :+1:

PR #29, opened 2024-07-23
Omit padding when editing structs

Cvelth:


lib/TypeNames/ModelToPTMLTypeHelpers.cpp
```diff
@@ -295,13 +295,33 @@ static void printDefinition(const model::EnumDefinition &E,
...
+static void printPadding(ptml::PTMLIndentedOstream &Header,
+                         ptml::PTMLCBuilder &B,
+                         uint64_t FieldOffset,
+                         uint64_t NextOffset,
+                         bool ForEditing) {
+  revng_assert(FieldOffset <= NextOffset);
+  if (FieldOffset == NextOffset)
+    return; // No padding is needed
+
+  if (ForEditing) {
```
aleclearmind, line 312: `ForEditing` is not great, there should be a boolean for each feature (e.g., `UseStartAt`), but at the current stage this would require rolling back the existing usages of `ForEditing` and probably exploding them in multiple arguments.

@pfez I'd suggest to let @Cvelth refactor this file so that we something more human like:

```c++
class Printer {
private:
  ptml::PTMLIndentedOstream &Header;
  ptml::PTMLCBuilder &B;
  bool UseStartAt;
  bool SomeOtherFeatureThatCurrentlyIsEmbeededInForEditing;
  Logger<> &Log;
  // ...

public:
  void printPadding(uint64_t FieldOffset, uint64_t NextOffset);
  void printDefinition(const model::StructDefinition &S);
  // ...
}
```

This is something we've been postponing way too long and it's starting to prevent us from having good designs.

lib/TypeNames/ModelToPTMLTypeHelpers.cpp
```diff
@@ -328,8 +328,12 @@ static void printDefinition(Logger<> &Log,
          << B.getKeyword(ptml::PTMLCBuilder::Keyword::Struct) << " "
          << B.getPackedAttribute() << " ";
 
-  if (ForEditing)
+  if (ForEditing) {
```
aleclearmind: Not just for editing, I'd like it in regular decompiled code too.

lib/ImportFromC/HeaderToModel.cpp
aleclearmind, line 1: @pfez can you review this?

lib/ImportFromC/HeaderToModel.cpp
```diff
@@ -892,8 +893,25 @@ bool DeclVisitor::handleStructType(const clang::RecordDecl *RD) {
     CurrentOffset += *Size;
   }
 
-  // TODO: Can this be calculated/fetched automatically?
-  Struct->Size() = CurrentOffset;
+  std::optional ExplicitSize = parseIntegerAnnotation(*Definition,
+                                                      SizeAnnotation);
+  if (not ExplicitSize.has_value()) {
+    revng_log(Log, "Explicit struct size tag is missing.");
+    // TODO: do we actually want to allow this through???
```
aleclearmind: While editing, the user should still be able to drop `_SIZE`, `_START_AT` and use padding fields.
Are we supporting both modes of operation?

share/revng/test/tests/analysis/ImportFromCAnalysis/backward_compatible_padding_struct.model.yml.ccode
aleclearmind, line 1: "backward compatible" hints at the fact that's depracted. It isn't, just a different way of doing things.

lib/TypeNames/ModelToPTMLTypeHelpers.cpp
pfez: I agree

lib/ImportFromC/HeaderToModel.cpp
pfez: Will do

lib/TypeNames/ModelToPTMLTypeHelpers.cpp
Cvelth: Sounds good!

lib/ImportFromC/HeaderToModel.cpp
Cvelth: This comment is specifically my doubt about whether we want size-less structs.

As for padding, yes, both modes can be used interchangeably. The only forbidden thing is `_START_AT` on an explicit padding field.

lib/TypeNames/ModelToPTMLTypeHelpers.cpp
aleclearmind: Nice, btw I'd be fine if this was an extra commit at the end.

share/revng/test/tests/analysis/ImportFromCAnalysis/backward_compatible_padding_struct.model.yml.ccode
Cvelth: `explicit_padding`? `array_padding`?

lib/TypeNames/ModelToPTMLTypeHelpers.cpp
Cvelth: Wait? Do you mean size? Or startAt? Or both?

aleclearmind: I meant `_CAN_CONTAIN_CODE`.

share/revng/test/tests/analysis/ImportFromCAnalysis/backward_compatible_padding_struct.model.yml.ccode
aleclearmind: 1. explicit_padding
2. use kebab case

share/revng/test/tests/analysis/ImportFromCAnalysis/rft-multiple-regs-for-return-val.model.yml.ccode
```diff
@@ -3,7 +3,7 @@
 //
 
 _ABI(raw_x86_64)
-struct return_value {
```
pfez, line 6: Actuall, size on the return value of a RFT doesn't make sense, because these aren't "real" types that live in memory, but they are only "bags of registers".

@aleclearmind any thought?

lib/ImportFromC/HeaderToModel.cpp
```diff
@@ -157,41 +158,61 @@ DeclVisitor::DeclVisitor(TupleTree<model::Binary> &Model,
...
 }
 
-// Parse location of parameter (`reg:` or `stack`).
-static std::optional<std::string> getLoc(llvm::StringRef Annotation) {
-  if (Annotation == "stack")
-    return Annotation.str();
+template<typename Type>
+static std::optional<uint64_t>
+parseIntegerAnnotation(const Type &Declaration, llvm::StringRef Prefix) {
+  std::optional<llvm::StringRef> Result = parseStringAnnotation(&Declaration,
```
pfez: I think this doesn't compile. Have you checked with `orc check branch`?

Cvelth: Fuck, I squashed the fix into the wrong commit (`HeaderToModel: add padding annotation support`). Thanks for noticing.

share/revng/test/tests/analysis/ImportFromCAnalysis/rft-multiple-regs-for-return-val.model.yml.ccode
Cvelth: Actually, considering that the consensus seems to be for `_SIZE(X)` to be optional, there is indeed no reason to add this here.

share/revng/test/tests/analysis/ImportFromCAnalysis/backward_compatible_padding_struct.model.yml.ccode
```diff
@@ -0,0 +1,9 @@
+//
+// This file is distributed under the MIT License. See LICENSE.md for details.
+//
+
+struct _PACKED my_struct {
+  uint8_t _padding_at_0[8];
+  uint64_t *field;
+  uint8_t _padding_at_8[8];
+};
```
pfez: I see that there is no REP for this PR. I'd like to know what are the design decisions about the following points:
1. Do we allow a mixed style of `_padding_at` and `_START_AT`?
2. If we don't allow mixed style, what happens at parse-time if the user uses mixed style?

Cvelth: We really should make one 😅 

1. Yep. Everything but `_START_AT` attached to a `_padding_at` field is allowed. `_START_AT` takes precedence.
2. Currently on any kind of non-clang error, it prints stuff to the log (which also seems to be broken under daemon) and then fails the analysis leaving model unchanged. That's the current behavour and I didn't touch it for now, since @aleclearmind promised me a better error reporting system at some point - and there's no use trying to push everything under the same rag until we have that.

Cvelth: By `_START_AT` takes precedence I mean that in cases like
```cpp
struct whatever {
  uint8_t _padding_at_0[16];
  _START_AT(8)
  uint64_t field;
};
```

explicit padding is pretty much ignored.

BTW, the only disallowed syntax is:
```cpp
struct whatever {
  _START_AT(8)
  uint8_t _padding_at_0[16];
  uint64_t field;
};
```
in which case we do indeed fail.

lib/ImportFromC/HeaderToModel.cpp
pfez: It's not clear to me if we allow both modes at the same time on a single struct, or it must be either one or the other.

lib/ImportFromC/HeaderToModel.cpp
```diff
@@ -26,6 +26,7 @@ static constexpr llvm::StringRef RawABIPrefix = "raw_";
 
 static constexpr llvm::StringRef ABIAnnotation = "abi:";
 static constexpr llvm::StringRef RegAnnotation = "reg:";
+static constexpr llvm::StringRef StackAnnotation = "stack";
```
pfez: Can you separate the rework of `stack` in a seprarate commit than the rework of parsing?

lib/ImportFromC/HeaderToModel.cpp
```diff
@@ -608,57 +597,52 @@ bool DeclVisitor::VisitFunctionDecl(const clang::FunctionDecl *FD) {
...
-          return false;
-        }
-
-        // TODO: Handle stack location.
-        if (*Loc == "stack") {
-          revng_log(Log, "We don't support parameters on stack for now");
+      std::optional Register = parseStringAnnotation(*ParamDecl, RegAnnotation);
+      if (not Register.has_value()) {
+        std::optional Stack = parseStringAnnotation(*ParamDecl,
+                                                    StackAnnotation);
```
pfez: I think in RFT we also allow only a single stack argument and it must have a pointer type in C. Isn't it?

lib/ImportFromC/HeaderToModel.cpp
Cvelth: It works one field at a time, so it would be somewhat ugly to explicitly decide "this struct only uses X".

lib/ImportFromC/HeaderToModel.cpp
Cvelth: Yep, sounds like a good assertion, thanks.

lib/ImportFromC/HeaderToModel.cpp
```diff
@@ -844,37 +814,25 @@ bool DeclVisitor::handleStructType(const clang::RecordDecl *RD) {
...
+        std::optional Stack = parseStringAnnotation(*Field, StackAnnotation);
+        if (Stack.has_value()) {
+          // TODO: Handle stack location.
+          revng_log(Log, "We don't support stack parameters in RFTs for now");
+          return false;
+        } else {
+          revng_log(Log,
+                    "Parameters must have either a register or a stack "
+                    "annotation.");
+          return false;
```
pfez: Also I think here we're assuming this is the return struct type of a RFT. The same restriction on having just one stack return value which is a pointer should apply I think.

pfez: BTW, do we have tests for these specific mistakes? I'd like to have tests of parsing types for which we expect failure.

Cvelth: Done! Please take a look at all the new tests :)

(I squashed them into the tests commit though)

lib/ImportFromC/HeaderToModel.cpp
Cvelth: It felt cheesy to extract information just so I can assert it, so I left a TODO instead. Please take a look at the new `HeaderToModel: leave some TODOs about RFT stack` commit.

lib/ImportFromC/HeaderToModel.cpp
Cvelth: I looked into it, but that's going to require non-trivial effort. Since the "pre-rework" version was pretty adamant about having single annotation per argument - and then dealing with how to parse it. The new version treats each annotation as its own thing - hence the need to differentiate the two.

lib/ImportFromC/HeaderToModel.cpp
Cvelth: Fixed!

share/revng/test/tests/analysis/ImportFromCAnalysis/rft-multiple-regs-for-return-val.model.yml.ccode
Cvelth: This is gone.

share/revng/test/tests/analysis/ImportFromCAnalysis/backward_compatible_padding_struct.model.yml.ccode
Cvelth: Done (and a bunch of new tests were added as well, feel free to take a fresh look at the commit).

lib/TypeNames/ModelToPTMLTypeHelpers.cpp
Cvelth: Done!

lib/TypeNames/ModelToPTMLTypeHelpers.cpp
Cvelth: Please take a look at #31 as an a bit of an overkill-y extension of this idea.

(I'm closing this thread, feel free to reopen it if you have further comments)

lib/ImportFromC/HeaderToModel.cpp
Cvelth: Closing this as Pietro confirmed the fact that he finished the review on mattermost.

aleclearmind:
This looks good to me. @pfez when you're done, mark as ready and merge.

lib/ImportFromC/HeaderToModel.cpp
pfez: Ok :shrug:

share/revng/test/tests/analysis/ImportFromCAnalysis/backward_compatible_padding_struct.model.yml.ccode
pfez: I don't think we have a test for the first thing you mentioned though, i.e. that `_START_AT` takes precedence. Can you please add it?

Also, I think that might be surprising, so make sure you `revng_log` it somewhere with message starting with `"WARNING: "`

lib/ImportFromC/HeaderToModel.cpp
pfez: To go back to your question @Cvelth: do we want size-less structs?

I think we do. The scenario I have in mind is: a user knows that some offsets in memory there's some data with a given type. They want to wrap it into a struct, and they don't want to be forced to manually compute it's size only to use it in the annotation. They want to just be able to tell `revng`: here's this struct, with these fields inside, and you figure out the minimum size needed to fit.

Cvelth: Yeah, I think we are on the same page as far as that is concerned. The check and TODO above are dropped already.

lib/ImportFromC/HeaderToModel.cpp
pfez: No offence, but it sounds silly. The very moment I read "TODO: don't forget to assert that it is always to assert" I think: "Why haven't you asserted then?".

So please, just assert it.

lib/ImportFromC/HeaderToModel.cpp
pfez: Cool thanks!

share/revng/test/tests/analysis/ImportFromCAnalysis/backward_compatible_padding_struct.model.yml.ccode
Cvelth: Done.

lib/ImportFromC/HeaderToModel.cpp
Cvelth: Done.

lib/ImportFromC/HeaderToModel.cpp
```diff
@@ -312,6 +312,13 @@ DeclVisitor::makePrimitive(const BuiltinType *UnderlyingBuiltin,
   return model::UpcastableType::empty();
 }
 
+static bool onlyContainsNumbers(llvm::StringRef Name) {
+  for (char Character : Name)
+    if (not llvm::StringRef("0123456789").contains(Character))
```
pfez:
```c++
if (not std::isdigit(Character))
```

lib/ImportFromC/HeaderToModel.cpp
```diff
@@ -587,15 +587,13 @@ bool DeclVisitor::VisitFunctionDecl(const clang::FunctionDecl *FD) {
                                                     StackAnnotation,
                                                     Errors);
         if (Stack.has_value()) {
-          // TODO: Handle stack location.
-          //
-          // TODO: Don't forget to assert that it is always a pointer.
-          Errors.emplace_back("revng: We don't support stack return values in "
-                              "RFTs for now.\n");
+          Errors.emplace_back("revng: RFTs must not have stack return values, "
+                              "use SPTAR instead.\n");
```
pfez: I don't think `SPTAR` is a common user-facing terminology.
Please expand the acronym to full text, so that users can understand what's `SPTAR`, and try to do the right thing to fix the problem.

pfez: Also, I see you're using the `revng` prefix in error messages.
It doesn't seem to me that we've ever agreed on putting that prefix.
If anything, I know @aleclearmind is pretty passionate about not putting `revng` prefixes: you're using `revng`, you kinda know the error messages come from `revng`.

The error messages are equally informative without the `revng` prefix. Just drop it please, here and in all the other few places around the codebase, that are all in `HeaderToModel.cpp`.

I know a few of these prefixes in error messages were there before your patchset, but the right thing to do is to remove them, not to spread them in even more places.

lib/ImportFromC/HeaderToModel.cpp
```diff
@@ -619,50 +619,71 @@ bool DeclVisitor::VisitFunctionDecl(const clang::FunctionDecl *FD) {
...
+        if (not Stack.has_value()) {
           Errors.emplace_back("revng: Arguments must have either a register or "
                               "a stack annotation.\n");
           return false;
+        } else {
+          if (I != N - 1) {
+            Errors.emplace_back("revng: Only the very last RFT argument is "
+                                "allowed to represent stack (which also means "
+                                "there can only be one! If you need to "
+                                "represent multiple values, use a struct).\n");
```
pfez: Please avoid parentheses or emphasis (!) in error messages.

lib/ImportFromC/HeaderToModel.cpp
```diff
@@ -619,50 +619,71 @@ bool DeclVisitor::VisitFunctionDecl(const clang::FunctionDecl *FD) {
...
+
+          if (not ParamType->isStruct()) {
+            Errors.emplace_back("revng: RFT stack argument must be a "
+                                "struct.\n");
+            return false;
+          }
+
+          if (not TheRawFunctionType.StackArgumentsType().isEmpty()) {
+            Errors.emplace_back("revng: Multiple stack arguments? How did this "
+                                "even happen?.\n");
```
pfez: Also, no irony in error messages.

share/revng/test/configuration/revng-c/import-from-c.yml
```diff
@@ -8,23 +8,25 @@ sources:
...
-      - struct-with-a-broken-field.model.yml
-      - struct-with-an-unsupported-field.model.yml
-      - struct-with-a-typo.model.yml
-      - struct-with-code.model.yml
-      - struct-with-conflicting-padding-information.model.yml
-      - struct-with-explicit-padding.model.yml
-      - struct-without-packed-attribute.model.yml
-      - struct-with-overlapping-fields.model.yml
-      - typedef.model.yml
-      - union.model.yml
```
pfez: Ah nice!

lib/ImportFromC/HeaderToModel.cpp
```diff
@@ -619,50 +619,71 @@ bool DeclVisitor::VisitFunctionDecl(const clang::FunctionDecl *FD) {
...
+          if (ParamDecl->getName() != "stack") {
+            Errors.emplace_back("revng: WARNING: stack argument name ('"
+                                + ParamDecl->getName().str()
+                                + "') was ignored.\n");
+          }
+        }
+      } else {
+        if (Stack.has_value()) {
+          Errors.emplace_back("revng: WARNING: _STACK annotation ignored "
+                              "because _REG takes precedence.\n");
```
pfez: I'm noting this down on this error message, but it's something that applies in general for errors reported by `HeaderToModel`, and possibly even more in general on errors on interactive user input.
So I'm also pulling @aleclearmind in the conversation because I'm sure he has opinions, and this is probably part of a broader design decision we'll eventually have to take.

Basically, the more I look at this, the more I'd like to have tests that test for expected failures.
But, with the current design of `Errors` as a `vector<string>` in `HeaderToModel`, this is only possible by checking for error messages, which means we're "pinning" error messages to their current form, which results in very brittle tests.

I believe the best thing to do would be to use `llvm::Error` instead of `std::string`s (see [here](https://llvm.org/docs/ProgrammersManual.html#recoverable-errors) and [here](https://llvm.org/doxygen/llvm_2Support_2Error_8h_source.html#l00089)).
If we use `llvm::Error` we can craft our own custom error classes for `HeaderToModel`. They can contain an error ID, and possibly other stuff. And by design of `llvm::Error` they must be convertible to a string.

Then we could write some unit-tests that check for error codes and avoid to "pin" error messages for testing expected failures.

I remember some arguing on `llvm::Error` and in particular on `llvm::ErrorList` in the past though.
I see `llvm::Error` is used throughout the `revng::pipeline` code extensively, but not `llvm::ErrorList`.
So maybe I'm missing something?
Can someone recap the main points of the old discussions on `llvm::Error` (and `llvm::ErrorList` if relevant)?
Any better ideas for the design?

Anyway, this is probably out of scope for this PR, so I'd put it in backlog, and for now we can either:
1. don't test for expected failures at all
2. test for expected failures by "pinning" error messages

@aleclearmind opinions? on this last point?

lib/ImportFromC/HeaderToModel.cpp
Cvelth: This is my way of contrasting `clang:` prefix. It would be weird for an imaginary error message list to looks something like
```
clang: blah-blah-from-clang
blah-blah-model-verification-failed
clang: not-sure-if-this-can-happen-but-who-knows-how-code-evolves
```

That being said, I'm not too attached, dropping the suffix is a mater of running a regex.

lib/ImportFromC/HeaderToModel.cpp
Cvelth: My two cents is that better error reporting is something the codebase severely lacks. I've been postponing this conversation because, in @aleclearmind's words, we have bigger fish to fry.

But we could definitely benefit from a more standardized error-warning reporting solution adopted through the whole pipeline.

It feel so bad, for example, to just write something to a logger when raw->cabi conversion fails - it's not like anyone but me is going to even look at that logger :D. If we had some centralized place to gather all warnings that user might want to see (bonus points for being able to group them by tiers / severity / where they are from, etc). Maybe even a log-like file appended to by the pipeline run that UI exposes or something. This needs some thought put into it, but we need something (maybe not anytime soon, but eventually - for sure. Good error reporting = less confused users). And I'd prefer for that something to be more handy that "vector of `llvm::Error`s".

-----------

As for the current state of `HeaderToModel`, we specifically use the brittle-testing @pfez mentioned:
1. Run the analysis on a given model with a given c-input.
2. If it succeeds, run `revng model compare` on the resulting model.
3. If it fails, compare its output against `error.txt` file included with the test (with stripped license header).

But, to be honest, it doesn't feel _that_ brittle because the tests itself are very granular _and_ they don't really depend on anything but the `HeaderToModel`, so the only way to break them - is to change `HeaderToModel`, in which case you need to rerun tests anyway - in which case we can easily catch the output change.

lib/ImportFromC/HeaderToModel.cpp
aleclearmind: I'd simply say "Cannot attach `_STACK` to return values in a RawFunctionDefinition". Replace `_STACK` with whatever annotation we're using.

In order to comment on the error messages, I'd need to see what's the current output.

lib/ImportFromC/HeaderToModel.cpp
pfez: Yeah in this specific situation I think that "pinning" error messages is acceptable until we don't do a proper redesign.

Speaking of which @Cvelth, given that you've added new failure modes, do all of them have associated tests with expected failure?

Cvelth: There's definitely one for every important failure, but if you prefer a more complete coverage, it shouldn't be hard to add more. Especially now that they are directory segregated.

aleclearmind: I think it's pretty clear that error reporting to user in edit types needs a lot of love.
But this is out of scope as of now.

For testing, clang does various stuff but they also hardcode error messages: https://github.com/llvm/llvm-project/blob/main/clang/docs/analyzer/developer-docs/DebugChecks.rst
It's not ideal, but I think it's fine.

I'm [adding a roadmap task about improving error reporting](https://rev.ng/gitlab/revng-company/release-roadmap/-/issues/140) (in edit types, but also in verify) for which we'll schedule the priority. But let's focus on the goal of this PR for now: adding support for `_START` and `_SIZE`.

lib/ImportFromC/HeaderToModel.cpp
Cvelth: 1. 👌.
2. That's a bit tricky, because for now the error reporting is just a vector of strings that eventually gets concatenated. Some of those strings come from clang (not sure if they can contain warnings (I didn't touch that part of the codebase that much, just turned it from "keep only the last error" to "just add every new error to the vector"), but I'd assume so), and some - from us (mostly when annotation don't make sense, or when the model stops verifying after the edit). I feel like there's a value in differentiating where the message is coming from, but not sure how much attention we should put into that for now. Once we have proper error handling, this whole thing will have to get redesigned anyway.

lib/ImportFromC/HeaderToModel.cpp
pfez: > if you prefer a more complete coverage, it shouldn't be hard to add more.

Yes please. I'd like at least one test that tests expected failure for each error message. Thanks!

> But let's focus on the goal of this PR for now

Sure, I stated it upfront. I just wanted to get the discussion started and agree on the best course of action now.
Thanks @aleclearmind

lib/ImportFromC/HeaderToModel.cpp
Cvelth: Done.

lib/ImportFromC/HeaderToModel.cpp
Cvelth: Done.

P. S. As for the current output samples, search for `expected-error.txt` files in the repo.

lib/ImportFromC/HeaderToModel.cpp
Cvelth: Simplified.

lib/ImportFromC/HeaderToModel.cpp
Cvelth: I turned this check into an assert, so the message is gone.

lib/ImportFromC/HeaderToModel.cpp
Cvelth: I made as many new tests as felt reasonable (see the caveats in the mattermost message for specifics). Please take a look at the very last commit.

share/revng/test/tests/analysis/ImportFromCAnalysis/cft-with-no-abi/expected-error.txt
```diff
@@ -0,0 +1,6 @@
+#
+# This file is distributed under the MIT License. See LICENSE.md for details.
+#
+
+import-from-c: Unable to parse the abi annotation.
```
pfez: ABI should be capital

share/revng/test/tests/analysis/ImportFromCAnalysis/enum-with-a-broken-annotation/expected-error.txt
```diff
@@ -0,0 +1,7 @@
+#
+# This file is distributed under the MIT License. See LICENSE.md for details.
+#
+
+import-from-c: An enum with a non-primitive underlying type.
```
pfez: Noting this here but it's a point that has general validity.

Shouldn't we include in the error message information that allows the user to identify the type that caused the error?
Remember that the user is trying to add a type from C (possibly in the UI) and would like an error message that is informative enough to locate the error and fix it.

Here for example there are 2 instances of the error.

I know from conversation on mattermost that this is just a quirk of enums somehow being parsed twice, but the point the main idea holds in general.
What if the type parsed by the user causes more than a single error? I expect the parser to fail on first error, but the error message should contain information to unambiguously understand what part of the input caused the problem.

share/revng/test/tests/analysis/ImportFromCAnalysis/enum-with-a-broken-annotation/expected-error.txt
```diff
@@ -0,0 +1,7 @@
...
+import-from-c: An enum with a non-primitive underlying type.
```
pfez: Also I think that it should be possible from the error message to discriminate between this case (the underlying type is not a valid type) from the case where the underlying type is a valid type but not a valid underlying type for an enum.

share/revng/test/tests/analysis/ImportFromCAnalysis/enum-without-an-annotation/expected-error.txt
```diff
@@ -0,0 +1,7 @@
+#
+# This file is distributed under the MIT License. See LICENSE.md for details.
+#
+
+import-from-c: Unable to parse the enum annotation.
```
pfez: What's the `enum annotation`?
What about `enum with invalid underlying type: unable to parse _ENUM_UNDERLYING annotation`?

share/revng/test/tests/analysis/ImportFromCAnalysis/rft-with-a-stack-return-value/expected-error.txt
```diff
@@ -0,0 +1,7 @@
+#
+# This file is distributed under the MIT License. See LICENSE.md for details.
+#
+
+import-from-c: Cannot attach _STACK to return values in a RawFunctionDefinition.
+import-from-c: Return value should have already been parsed.
```
pfez: Again on the wording of the error message. What about `Missing definition of return type`?

share/revng/test/tests/analysis/ImportFromCAnalysis/rft-with-an-unannotated-argument/expected-error.txt
```diff
@@ -0,0 +1,6 @@
+#
+# This file is distributed under the MIT License. See LICENSE.md for details.
+#
+
+import-from-c: Arguments must have either a register or a stack annotation.
```
pfez: I'd like the syntax of the annotations to be spelled out in the error messages. To help the user conform.

This applies to all errors about annotations in general, not just here.

share/revng/test/tests/analysis/ImportFromCAnalysis/rft-with-an-unknown-architecture/expected-error.txt
```diff
@@ -0,0 +1,6 @@
+#
+# This file is distributed under the MIT License. See LICENSE.md for details.
+#
+
+import-from-c: Invalid raw abi architecture.
```
pfez: Capital ABI

share/revng/test/tests/analysis/ImportFromCAnalysis/rft-with-an-unknown-register/expected-error.txt
```diff
@@ -0,0 +1,7 @@
+#
+# This file is distributed under the MIT License. See LICENSE.md for details.
+#
+
+import-from-c: Unsupported register location.
```
pfez: `Unsupported register location in _REG annotation.`

BONUS

`Valid values are: ` and then you print the valid values we accept for the given ABI

pfez: The latter shouldn't be part of the test though. It makes it much more brittle.

share/revng/test/tests/analysis/ImportFromCAnalysis/struct-with-a-broken-annotation/expected-error.txt
```diff
@@ -0,0 +1,6 @@
+#
+# This file is distributed under the MIT License. See LICENSE.md for details.
+#
+
+import-from-c: Ignoring non-integer value of an integer annotation: 'sixteen'.
```
pfez: `of an integer _START_AT annotation`

share/revng/test/tests/analysis/ImportFromCAnalysis/struct-with-a-rare-type/expected-error.txt
```diff
@@ -0,0 +1,7 @@
+#
+# This file is distributed under the MIT License. See LICENSE.md for details.
+#
+
+import-from-c: Unsupported QualType.
```
pfez: What's a `QualType`? I know it's `clang`'s stuff but this is not informative for users. Perhaps print the type name?

share/revng/test/tests/analysis/ImportFromCAnalysis/struct-with-conflicting-annotations/expected-error.txt
```diff
@@ -0,0 +1,7 @@
+#
+# This file is distributed under the MIT License. See LICENSE.md for details.
+#
+
+import-from-c: Multiple conflicting annotation values found: '8' and '16'.
```
pfez: `Multiple conflicting _START_AT annotations`

pfez (changes requested):
This is really going in a good direction. New tests are very good.

We only need a last iteration on the specific wording I think, and then we're done

lib/ImportFromC/HeaderToModel.cpp
Cvelth: Closing this as you did indeed take a look (thanks).

share/revng/test/tests/analysis/ImportFromCAnalysis/cft-with-no-abi/expected-error.txt
Cvelth: Done, elsewhere too.

share/revng/test/tests/analysis/ImportFromCAnalysis/enum-with-a-broken-annotation/expected-error.txt
Cvelth: Doing so reliably would definitely be non-trivial. I did my best in some of the more egregious cases (see updated tests for details), but it's not, by any means, anywhere near being usable. But I guess that's why `@aleclearmind` scheduled that error-message clean up task.

share/revng/test/tests/analysis/ImportFromCAnalysis/enum-with-a-broken-annotation/expected-error.txt
Cvelth: Done.

share/revng/test/tests/analysis/ImportFromCAnalysis/enum-without-an-annotation/expected-error.txt
Cvelth: Done.

share/revng/test/tests/analysis/ImportFromCAnalysis/rft-with-a-stack-return-value/expected-error.txt
Cvelth: I went with `Unable to parse the type of the return value.`

share/revng/test/tests/analysis/ImportFromCAnalysis/rft-with-an-unannotated-argument/expected-error.txt
Cvelth: Done (where reasonable).

share/revng/test/tests/analysis/ImportFromCAnalysis/rft-with-an-unknown-architecture/expected-error.txt
Cvelth: Done.

share/revng/test/tests/analysis/ImportFromCAnalysis/rft-with-an-unknown-register/expected-error.txt
Cvelth: I disagree that we should encode all the register names in here. But even if we do, that needs to be more systematic and done for all the annotations. And some of them (say abi) are not going to be capped to a specific list (since adding an abi is, eventually, going to be as easy as adding a new file to a specified directory). Which makes it even more awkward. But I guess we can further discuss this as the part the error clean up task `@aleclearmind` scheduled for this.

share/revng/test/tests/analysis/ImportFromCAnalysis/struct-with-a-broken-annotation/expected-error.txt
Cvelth: Sadly, this is not possible. The code for parsing these is shared, and it's definitely not easy (if even possible) to extract macro names like `_START_AT` from clang. Annotations themselves (which macros unwrap into) is a lot easier, so I used this one for now, even though I'm not sure how useful those would be for the user.

Eventually, we might want to maintain are mapping between the two for the users' sake, but even the thought makes me kind of sad.

share/revng/test/tests/analysis/ImportFromCAnalysis/struct-with-a-rare-type/expected-error.txt
Cvelth: That was a catch all error for failing to parse whatever clang spitted out. I went with:
```
import-from-c: Unsupported type.
import-from-c: This error indicates that the type system clang emitted contains something we either decided not to allow in the model OR do not support yet.
import-from-c failed: Unable to parse the type of struct field #0.
```

share/revng/test/tests/analysis/ImportFromCAnalysis/struct-with-conflicting-annotations/expected-error.txt
Cvelth: This thread is actually the same as the [above one](https://github.com/revng/revng-c/pull/29#discussion_r1734868434), so I'll close it so that the discussion doesn't get split.

lib/ImportFromC/HeaderToModel.cpp
```diff
@@ -1010,17 +1086,20 @@ bool DeclVisitor::handleUnionType(const clang::RecordDecl *RD) {
   uint64_t CurrentIndex = 0;
   for (const FieldDecl *Field : Definition->fields()) {
     if (Field->isInvalidDecl()) {
-      Errors.emplace_back("import-from-c: Invalid declaration for a union "
-                          "field.\n");
+      Errors.emplace_back("import-from-c failed: The declaration of the union "
+                          "field #'"
+                          + std::to_string(Union->Fields().size())
```
pfez, line 1057: `.size()`? :thinking:

pfez: Can't you tell the field number?

pfez: Aaah we're parsing, so this is fine because `size()` is also the index of the current field being parsed. nice.

lib/ImportFromC/HeaderToModel.cpp
```diff
@@ -639,43 +672,62 @@ bool DeclVisitor::VisitFunctionDecl(const clang::FunctionDecl *FD) {
...
+                                  "struct.\n");
             return false;
           }
 
           if (not ParamType->isStruct()) {
-            Errors.emplace_back("import-from-c: RFT stack argument must be a "
-                                "struct.\n");
+            // TODO: should we allow non-structs here? The only downside is that
+            //       it would mean transparently adding an extra wrapping struct
+            //       to the model, which doesn't sounds too bad to me.
```
pfez: In principle we could, but for now there's no advantage, and lots of places in the code still assume it's a struct.

lib/ImportFromC/HeaderToModel.cpp
```diff
@@ -626,8 +658,9 @@ bool DeclVisitor::VisitFunctionDecl(const clang::FunctionDecl *FD) {
       auto QT = ParamDecl->getType();
       model::UpcastableType ParamType = getModelTypeForClangType(QT);
       if (not ParamType) {
-        Errors.emplace_back("import-from-c: Unsupported type for a raw "
-                            "function argument.\n");
+        Errors.emplace_back("import-from-c failed: Unable to parse the type "
+                            "the argument #"
```
pfez, line 644: Missing `of`

lib/ImportFromC/HeaderToModel.cpp
```diff
@@ -546,8 +572,9 @@ bool DeclVisitor::VisitFunctionDecl(const clang::FunctionDecl *FD) {
       auto QT = FD->getParamDecl(I)->getType();
       model::UpcastableType ParamType = getModelTypeForClangType(QT);
       if (not ParamType) {
-        Errors.emplace_back("import-from-c: Unsupported type for a function "
-                            "argument.\n");
+        Errors.emplace_back("import-from-c failed: Unable to parse the type "
+                            "the argument #"
```
pfez, line 562: Missing `of`

lib/ImportFromC/HeaderToModel.cpp
```diff
@@ -485,15 +505,19 @@ DeclVisitor::getModelTypeForClangType(const QualType &QT) {
       else if (auto Rw = makeTypeByNameOrID<model::RawFunctionDefinition>(Name))
         R = std::move(Rw);
       else
-        Errors.emplace_back("import-from-c: Couldn't find function type in the "
-                            "model.\n");
+        Errors.emplace_back("import-from-c: Unknown typedef: '" + Name.str()
+                            + "'. Is it present in the model?\n");
```
pfez: And what if not? See other places too.

Do you know that it's not present in the model? If so, just state it instead of asking questions.

If you ask questions you should also provide further directions depending on the answer to the questions, and the error message becomes a walkthrough.

I'd drop the question from the error message.

share/revng/test/tests/analysis/ImportFromCAnalysis/struct-with-conflicting-annotations/expected-error.txt
```diff
@@ -2,6 +2,5 @@
 # This file is distributed under the MIT License. See LICENSE.md for details.
 #
 
-import-from-c: Multiple conflicting annotation values found: '8' and '16'.
-
+import-from-c:7:16: Multiple conflicting values ('8' and '16') were found for the 'field_start_offset:' annotation.
```
pfez: what's `field_start_offset` annotation? that's not something the user ever sees. what they see is `_START_AT` and that's also what they should be seeing here.

lib/ImportFromC/HeaderToModel.cpp
Cvelth: Yep, we're inserting them one at a time. If we go with indexing from 0, the number of already processed fields is our index.

lib/ImportFromC/HeaderToModel.cpp
Cvelth: No, that's not what I meant!

I meant allowing the users to specify a non-struct, and then we transparently make a new struct for them to wrap whatever type they gave.

lib/ImportFromC/HeaderToModel.cpp
Cvelth: No, that was just a cheeky way of saying "hey, check your model, you probably forgot to add it".

But yeah, I'm fine with just `Unknown typedef: 'name_with_a_typo'.`.

share/revng/test/tests/analysis/ImportFromCAnalysis/struct-with-conflicting-annotations/expected-error.txt
Cvelth: To not split the discussion, I'll just link the other related thread here: https://github.com/revng/revng-c/pull/29#discussion_r1740754501

lib/ImportFromC/HeaderToModel.cpp
pfez: neat

lib/ImportFromC/HeaderToModel.cpp
pfez: no, I'd like to minimize surprise for users. and a magical new wrapper struct is surprising. let's leave it like it is for now

share/revng/test/tests/analysis/ImportFromCAnalysis/enum-with-a-broken-annotation/expected-error.txt
pfez: makes sense. what you've done is more than enough for now

lib/ImportFromC/HeaderToModel.cpp
Cvelth: Gone!

lib/ImportFromC/HeaderToModel.cpp
Cvelth: Done.

lib/ImportFromC/HeaderToModel.cpp
Cvelth: Done.

lib/ImportFromC/HeaderToModel.cpp
Cvelth: Done.

share/revng/test/tests/analysis/ImportFromCAnalysis/rft-with-a-stack-return-value/expected-error.txt
pfez: :+1:

share/revng/test/tests/analysis/ImportFromCAnalysis/struct-with-a-rare-type/expected-error.txt
pfez: Ok, but let's this less verbose and compact everything in a single line
```
import-from-c: unsupported type: C type provided for struct field #0 is not supported in the model
```

share/revng/test/tests/analysis/ImportFromCAnalysis/rft-with-an-unknown-register/expected-error.txt
pfez: It's true that when we support user-defined ABIs it's gonna be harder to enumerate the list of valid registers for each ABI, but I expect that to be perfectly doable at runtime. If there's a user-defined ABI, that will be picked up and will be dynamically available to this parsing routines too.

Anyway, let's leave the register names out of the picture for now. We'll deal with them when we properly rehaul error messages in the future.

share/revng/test/tests/analysis/ImportFromCAnalysis/struct-with-a-rare-type/expected-error.txt
Cvelth: Wait, don't forget that the real error
```
import-from-c failed: Unable to parse the type of struct field #0.
```

is emitted elsewhere, so it would be somewhat hard to get rid of it.

What about something like:
```
import-from-c: The type is cannot be represented in the model.
import-from-c failed: Unable to parse the type of struct field #0.
```
?

share/revng/test/tests/analysis/ImportFromCAnalysis/struct-with-conflicting-annotations/expected-error.txt
pfez: I'm leaving this open, but I'll continue answering only the other one.

Cvelth: Sure, but consider that however we decide to proceed here was always going to affect every single annotation (with a value) anyway.

share/revng/test/tests/analysis/ImportFromCAnalysis/struct-with-a-broken-annotation/expected-error.txt
pfez: Ah damn, macros!!!

I understand that they are a pain, but at the moment, if I put myself in a user's shoes, what I see it the following situation:
- in C for specifying types I'm instructed to use macros such as `_START_AT` and others (like `_REG`, `_STACK`)
- in most error messages I can just see the macros directly (again `_STACK`, `_REG`)
- I never see the `field_start_offset` anywhere in the C code generated by `revng` and I don't know how to use it
- I get error messages containing `field_start_offset` when parsing my C code, whereas I never used `field_start_offset` at all

This is very confusing.

So, I appreciate the effort to make the code generic, but this currently degrades the UX.

The mapping for the users' sake at this point is important, and it's very easy to implement. So please do it.
You don't need to care about making it extensible by user-defined annotations yet. Just go with a simple design. Even a `static const` global map is fine.

share/revng/test/tests/analysis/ImportFromCAnalysis/struct-with-a-rare-type/expected-error.txt
pfez: > that the real error is emitted elsewhere

Ah right.

I like your proposal then. `is cannot ` is a typo though.
Besides that, I approve this

share/revng/test/tests/analysis/ImportFromCAnalysis/struct-with-conflicting-annotations/expected-error.txt
pfez: yeah I'm aware

include/revng-c/Support/Annotations.h
```diff
@@ -0,0 +1,135 @@
...
+    auto Result = std::ranges::find_if(StaticAnnotations, [](auto &&A) {
+      return *Macro == A.Macro;
+    });
+    if (Result == StaticAnnotations.end())
+      return std::nullopt;
+    return *Result;
+  }
+
+public:
+  // NOTE: PTML-tags are omitted for now, but eventually they should be included
```
pfez: `s/NOTE/TODO` everywhere.

include/revng-c/Support/PTMLC.h
```diff
@@ -419,32 +420,19 @@ class PTMLCBuilder : public ptml::PTMLBuilder {
...
+  std::string getAnnotation(uint64_t Value) const {
+    return Attributes.getReference<Macro>(std::to_string(Value));
   }
-
-  std::string getRegisterAnnotation(const llvm::StringRef RegisterName) const {
-    return "_REG(" + RegisterName.str() + ")";
+  template<ConstexprString Macro>
+  std::string getAttribute() const {
+    return Attributes.getReference<Macro>();
   }
```
pfez: These don't really need to live in `PTMLCBuilder`, do they? They seem to belong to the new `Annotation.h` header.

include/revng-c/Support/Annotations.h
```diff
@@ -0,0 +1,135 @@
...
+  template<ConstexprString Macro>
+  std::string getReference() const {
```
pfez: Why `getReference`? This makes a copy. Just `get` I guess.

pfez: Same below.

include/revng-c/Support/Annotations.h
pfez: Also, can you be more precise about what you mean? Not necessarily in the comment yet, but what do you have in mind for PTML tags?

include/revng-c/Support/Annotations.h
```diff
@@ -0,0 +1,135 @@
...
+  std::string_view Value;
+  bool IsReal = false;
+};
+
+// Annotation is an attribute that can also encode a value
+// (for example an abi name).
+struct Annotation {
+  std::string_view Macro;
+  std::string_view Prefix;
+  std::string_view Suffix;
```
pfez: It seems to me that this is unused.

Not even the `getDefinitions` function uses this.

pfez: Let's just drop this for now.

pfez (changes requested):
Just the final touches are missing.

I was initially not convinced by using this approach based on `ConstexprString`, because you end up having the string literals sprinkled a lot throughout the codebase.
But for now I think is fine.

The only thing still bugging me is the double parsing of `enum`s. I'm checking a couple of things to see if we can avoid that.

share/revng/test/tests/analysis/ImportFromCAnalysis/struct-with-a-broken-annotation/expected-error.txt
pfez: Closing this as done. Thank you.

share/revng/test/tests/analysis/ImportFromCAnalysis/struct-with-a-rare-type/expected-error.txt
Cvelth: Done (forgot to close).

include/revng-c/Support/Annotations.h
Cvelth: The same thing we need to do with the primitives header - generate it wrapping it into a bunch of `definition` tags, and then wrap all the usages into `reference` tags. Which makes all of these things magically control-clickable in the ui :enterprise:

include/revng-c/Support/PTMLC.h
Cvelth: To be fair, yeah, these are noop wrappers. I'm a bit hesitant whether we want to make annotation things ptml-aware or not. If not, these are going to be tasked with wrapping stuff into ptml tags.

include/revng-c/Support/Annotations.h
Cvelth: Same thing as elsewhere, I'm still unsure whether we want to wrap stuff in here, or in the builder. Once we get to wrapping. As such, I just left as many dangling ends as I could so we can decide later :D

include/revng-c/Support/Annotations.h
Cvelth: Ah, whoops, thanks for noticing. It was a left-over. In an earlier implementation, prefix also contained `annotate("`, which required `")` suffix. But that proven to bring more headaches that it solved. But it seems i forgot to remove the suffix :D

include/revng-c/Support/Annotations.h
Cvelth: Closing this as `s/NOTE/TODO` is done, and I started a mattermost discussion for the second part.

include/revng-c/Support/Annotations.h
Cvelth: Gone.

include/revng-c/Support/PTMLC.h
Cvelth: Gone.

include/revng-c/Support/Annotations.h
Cvelth: I just merged `getAttribute` and `getAnnotation` from the builder into this instead.

pfez: Looks good!

PR #30, opened 2024-08-06
DetectStackSize: handle counted loops

aleclearmind:
See companion [revng-qa PR](https://github.com/revng/revng-qa/pull/28).

### DIFF
--- a/include/revng-c/PromoteStackPointer/ComputeStackAccessesBoundsPass.h
+++ b/include/revng-c/PromoteStackPointer/ComputeStackAccessesBoundsPass.h
@@ -7,13 +7,13 @@
 #include "llvm/Pass.h"
 
 /// Enrich StackOffsetMarker-tagged calls with LazyValueInfo boundaries info
-struct ComputeStackAccessesBoundsPass : public llvm::ModulePass {
+struct ComputeStackAccessesBoundsPass : public llvm::FunctionPass {
 public:
   static char ID;
 
-  ComputeStackAccessesBoundsPass() : llvm::ModulePass(ID) {}
+  ComputeStackAccessesBoundsPass() : llvm::FunctionPass(ID) {}
 
   void getAnalysisUsage(llvm::AnalysisUsage &AU) const override;
 
-  bool runOnModule(llvm::Module &M) override;
+  bool runOnFunction(llvm::Function &F) override;
 };
--- a/lib/PromoteStackPointer/ComputeStackAccessesBoundsPass.cpp
+++ b/lib/PromoteStackPointer/ComputeStackAccessesBoundsPass.cpp
@@ -2,11 +2,23 @@
 // This file is distributed under the MIT License. See LICENSE.md for details.
 //
 
+#include <optional>
+#include <utility>
+
+#include "llvm/ADT/STLExtras.h"
+#include "llvm/ADT/SmallVector.h"
 #include "llvm/Analysis/LazyValueInfo.h"
+#include "llvm/Analysis/LoopInfo.h"
+#include "llvm/Analysis/PostDominators.h"
+#include "llvm/Analysis/ScalarEvolution.h"
+#include "llvm/Analysis/ScalarEvolutionExpressions.h"
+#include "llvm/IR/BasicBlock.h"
 #include "llvm/IR/ConstantRange.h"
+#include "llvm/IR/Dominators.h"
 #include "llvm/IR/Instructions.h"
 
 #include "revng/Model/IRHelpers.h"
+#include "revng/Support/IRHelpers.h"
 
 #include "revng-c/PromoteStackPointer/ComputeStackAccessesBoundsPass.h"
 #include "revng-c/PromoteStackPointer/InstrumentStackAccessesPass.h"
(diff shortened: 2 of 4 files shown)


lib/PromoteStackPointer/ComputeStackAccessesBoundsPass.cpp
```diff
@@ -2,58 +2,155 @@
...
+static ConstantRange
+getSCEVBoundaries(ScalarEvolution &SE, PostDominatorTree &PDT, Instruction *I) {
+  const auto *SCEV = SE.getSCEV(I);
+  auto FullSet = ConstantRange::getFull(SCEV->getType()->getIntegerBitWidth());
+
+  auto *AddRec = dyn_cast<SCEVAddRecExpr>(SCEV);
+  if (AddRec == nullptr)
+    return FullSet;
+
+  // Only consider loop ins loop simplify form
```
pfez: Typo `ins`

lib/PromoteStackPointer/ComputeStackAccessesBoundsPass.cpp
```diff
@@ -2,58 +2,155 @@
...
+  BasicBlock *Location = I->getParent();
+  if (not PDT.dominates(Location, Loop->getHeader()))
+    return FullSet;
+
+  // Collect operands
+  // TODO: handle multiple AddRecs
+  const auto *Start = AddRec->getStart();
+  auto *BackedgeTakenCount = SE.getBackedgeTakenCount(Loop);
+  const auto *Stride = AddRec->getStepRecurrence(SE);
+  APInt One(Start->getType()->getIntegerBitWidth(), 1);
```
pfez: Drop this and just use `SE.getOne(Start->getType())` in its only use.

lib/PromoteStackPointer/ComputeStackAccessesBoundsPass.cpp
```diff
@@ -2,58 +2,155 @@
...
+  // Ensure bounds could be computed and are constant
+  if (not areAll<SCEVConstant>(std::vector{ Start, End }))
+    return FullSet;
+
+  APInt StartValue = cast<SCEVConstant>(Start)->getAPInt();
+  APInt EndValue = cast<SCEVConstant>(End)->getAPInt();
+
+  // TODO: handle backward loops
+  if (EndValue.slt(StartValue))
+    return FullSet;
```
pfez: Can't you just `return ConstantRange(EndValue-1, StartValue-1)` here?

This should be correct considering that IIRC `ConstantRange` includes the lower but not the higher.

lib/PromoteStackPointer/ComputeStackAccessesBoundsPass.cpp
```diff
@@ -2,58 +2,155 @@
...
     auto *FunctionType = StackOffsetFunction.getFunctionType();
     auto *DifferenceType = cast<IntegerType>(FunctionType->getParamType(1));
-    for (CallBase *Call : callers(&StackOffsetFunction)) {
-      Function &F = *Call->getParent()->getParent();
-      LazyValueInfo &LVI = getAnalysis<LazyValueInfoWrapperPass>(F).getLVI();
+
+    for (CallBase *Call : callersIn(&StackOffsetFunction, &F)) {
       auto *Undef = UndefValue::get(DifferenceType);
 
       // Identify lower bound of the lower bound
```
pfez: This comment is stale.

lib/PromoteStackPointer/ComputeStackAccessesBoundsPass.cpp
```diff
@@ -2,58 +2,155 @@
...
+        auto *V = Operand.get();
+        const APInt *Bound = nullptr;
+
+        // Get range from LVI
+        // TODO: using LVI is inaccurate (unlike using SCEV, the way we use it)
+        const auto &BoundRange = LVI.getConstantRange(V, cast<CallInst>(User));
+        if (not BoundRange.isFullSet()) {
+          Bound = Lower ? &BoundRange.getLower() : &BoundRange.getUpper();
+        }
+
```
pfez: This can be moved in the next `if`

lib/PromoteStackPointer/ComputeStackAccessesBoundsPass.cpp
```diff
@@ -2,58 +2,155 @@
...
-        UpperBound = Undef;
-      }
-      Call->setArgOperand(2, UpperBound);
+      auto ReplaceUseWithBound = [&](Use &Operand, bool Lower) {
+        auto *User = Operand.getUser();
+        auto *V = Operand.get();
+        const APInt *Bound = nullptr;
+
+        // Get range from LVI
+        // TODO: using LVI is inaccurate (unlike using SCEV, the way we use it)
```
pfez: Can you elaborate? This isn't obvious for someone who doesn't know the internals of both LVI and SCEV.

lib/PromoteStackPointer/ComputeStackAccessesBoundsPass.cpp
```diff
@@ -2,58 +2,155 @@
...
+              Bound = &SCEVBound;
+            }
+          } else {
+            SCEVBound = SCEVRange.getUpper();
+            if (not SCEVBound.isMaxSignedValue()
+                and (Bound == nullptr or SCEVBound.sle(*Bound))) {
+              Bound = &SCEVBound;
+            }
+          }
+        }
```
pfez: It's weird because you update the AVI min only if the AVI min doesn't exist or the SCEV min is higher, and the AVI max only if the AVI mas doesn't exist or the SCEV min is lower.

At first it seems an error, but then I this has to do with the `TODO` above, where you say that LVI is inaccurate.
Overall this whole thing about inaccuracy can use some better explanation in comments.

pfez: Also, it seems weird that you use `sgt` for `Lower` and `sle` (instead of `slt`) for the other case. I know that in `llvm::ConstantRange` the upper value is considered not included in the range and the lower is. But here we're comparing upper values with upper values, so it seems to me that the same comparison should apply.

share/revng/test/tests/analysis/DetectStackSize/dynamic_native/memcpy-from-stack-arguments.c.model.yml
```diff
@@ -0,0 +1,16 @@
+---
```
pfez, line 5: I'm putting this here because I don't know where else to put it.

Looking at my comment above on using `slt` vs `sle`, if it turns out I am right and it was a bug, I'd like to have a test for that, so that whenever we have to change this it doesn't slip back in.

include/revng-c/PromoteStackPointer/ComputeStackAccessesBoundsPass.h
```diff
@@ -7,13 +7,13 @@
 #include "llvm/Pass.h"
 
 /// Enrich StackOffsetMarker-tagged calls with LazyValueInfo boundaries info
-struct ComputeStackAccessesBoundsPass : public llvm::ModulePass {
+struct ComputeStackAccessesBoundsPass : public llvm::FunctionPass {
```
pfez, line 10: Finally!!!

pfez (changes requested):

lib/PromoteStackPointer/ComputeStackAccessesBoundsPass.cpp
aleclearmind: LVI gives you an upper bound, not an exact bound. I'll add a comment.

lib/PromoteStackPointer/ComputeStackAccessesBoundsPass.cpp
pfez: Also, you might want to use `llvm::APIntOps::smax(Bound, SCEVBound)` and `smin` instead of doing the weird dance with `sle`/`slt` and `sgt`. I think using them the `APIntOps` would have prevented the error in the first place.

lib/PromoteStackPointer/ComputeStackAccessesBoundsPass.cpp
aleclearmind: Maybe I could, but I think I'm assuming FullSet means "know nothing". And if `EndValue.slt(StartValue)` it's not like a regular counted loop.

If I do this, we should probably find some threshold to say "if the range is bigger than X", bail out.
Basically I'm trying to handle only the very nice cases, but  see that's not perfect.

What do you suggest?

lib/PromoteStackPointer/ComputeStackAccessesBoundsPass.cpp
aleclearmind: No, idea is that we get bounds for `ConstantInt`s from LVI.

lib/PromoteStackPointer/ComputeStackAccessesBoundsPass.cpp
aleclearmind: > AVI min

Note that AVI is not involved at all here. You probably meant LVI.

> At first it seems an error

Why? I've two candidate lower bounds and I'm looking for the strictest among the two. The strictest is the highest one. Viceversa for the upper bound.

> Also, it seems weird that you use sgt for Lower and sle (instead of slt) for the other case

I fixed it for coherency purposes, but it's irrelevant. The case where both bounds are equal can both lead to keep the old value or update to the new one and makes no difference: they are the same.

> llvm::APIntOps::smax

Didn't know about that, I can use it but I just like the idea of having a code path that is taken only if we successfully found a stricter bound due to SCEV. For instance I can breakpoint there, add some logging.
But no strong opinions.

lib/PromoteStackPointer/ComputeStackAccessesBoundsPass.cpp
aleclearmind: I've added a more verbose comment.

lib/PromoteStackPointer/ComputeStackAccessesBoundsPass.cpp
pfez: > And if EndValue.slt(StartValue) it's not like a regular counted loop.

Yeah. I was thinking it must be a loop starting from a high value and iterating in reverse, but perhaps it could be an overflow? Is that what you're concerned of?

lib/PromoteStackPointer/ComputeStackAccessesBoundsPass.cpp
pfez: Not sure what you mean.

My point is that you can do this chunk of computation only on the paths where `V` is an `Instruction`, so you can push this inside the next `if (auto *I = dyn_cast<Instruction>(V))`. It will save you a few (not many) computations. But I have no idea if these computations might end up being costly in some weird corner case.

lib/PromoteStackPointer/ComputeStackAccessesBoundsPass.cpp
pfez: Thanks

lib/PromoteStackPointer/ComputeStackAccessesBoundsPass.cpp
pfez: > You probably meant LVI.

Yes.

> At first it seems an error

If you don't know what the code is trying to do, with the funny dance with LVI and SCEV. With the new expanded comment about upper bounds it's less confusing now.

> I fixed it for coherency purposes, but it's irrelevant. 

Right, but it's less confusing now. Thanks.

> Didn't know about that, I can use it but I just like the idea of having a code path that is taken only if we successfully found a stricter bound due to SCEV. For instance I can breakpoint there, add some logging.

Fair point. Let's leave it like this.

lib/PromoteStackPointer/ComputeStackAccessesBoundsPass.cpp
aleclearmind: Yes.

lib/PromoteStackPointer/ComputeStackAccessesBoundsPass.cpp
aleclearmind: No, I specifically want to handle the situation where `V` is a `ConstantInt` (and, therefore, not an `Instruction`) with LVI.

If `V` is a `ConstantInt`, LVI will gives us an appropriate `ConstantRange`. Pushing these lines within the `if` means not handling `ConstantInt`s, which is not what we want.

(fixed a typo in my previous comment)

pfez: :+1:

lib/PromoteStackPointer/ComputeStackAccessesBoundsPass.cpp
pfez: I'm checking how `ScalarEvolution` handles overflows/underflows

pfez: So, to do it properly you'd have to deal with `NSW` flags.

But basically you can manually check no overflow/underflow happens before calling `getAddExpr`.
If no wrap happens you can do like I suggested above, because no wrap is guaranteed.
Otherwise you return `FullSet` on wrap and you're done.

How does that sound?

aleclearmind: I'm a bit confused, what are we trying to achieve? We're discussing implementation details but we're not stating what's the goal.

Are we trying to support backward loops? If so, I can look into it and add a test, but I think we discussed this and said it was only supposed to handle simple `memcpy`s.

The overflow problem can happen also with situation where `End > Start` (if the stride is negative). So I can either bail out in case the stride is negative (and avoid the overflow issue), or I can handle backward loops too and (if necessary) figure out how overflowing SCEV work and add a test for backward loops. However, if you ask for an extra unit of work that we decided to leave out, please state it.

pfez: It seems to me that handling negative strides is marginally more complex, so I think it's worth it to handle it now while we're fresh on the topic.

pfez: And I understand that you have to handle both overflow with positive stride and underflow with negative stride. But we have facilities, like `OverflowSafeInt`, that should make it easier.

lib/PromoteStackPointer/ComputeStackAccessesBoundsPass.cpp
```diff
@@ -2,58 +2,188 @@
...
+    llvm::SmallVector<const llvm::SCEV *, 4> Addends = {
+      Start,
+      SE.getMulExpr(Stride, BackedgeTakenCount),
+      SE.getOne(Start->getType())
+    };
+    End = SE.getAddExpr(Addends);
+  } else {
+    End = SE.getAddExpr(Start, SE.getOne(Start->getType()));
+    Start = SE.getAddExpr(Start, SE.getMulExpr(Stride, BackedgeTakenCount));
+  }
```
aleclearmind, line 84: @pfez We now also handle backward loops, I added a test too. See `backward_memcpy_from_stack_arguments` in this PR and in the [revng-qa's one](https://github.com/revng/revng-qa/pull/28/files).

See if it looks good to you.

lib/PromoteStackPointer/ComputeStackAccessesBoundsPass.cpp
aleclearmind: I now handle backward loops, added another discussion.

lib/PromoteStackPointer/ComputeStackAccessesBoundsPass.cpp
pfez: Yeah great thanks!

PR #31, opened 2024-08-07
Introduce `ptml::CTypeBuilder`

Cvelth:
Sister PR: https://github.com/revng/revng/pull/401


aleclearmind:
@pfez I'm assigning this to you. Reminder: only the last commit of this PR should be reviewed. The others are part of another PR.

include/revng-c/TypeNames/PTMLCTypeBuilder.h
```diff
@@ -0,0 +1,245 @@
...
+  /// When set to true, the types that are only ever depended on by a single
+  /// other type (for example, a nested struct), are printed inside the parent
+  /// type definition instead of separately.
+  bool EnableTypeInlining = true;
+
+  /// When set to true, function stack frame types are printed inside the struct
+  /// body as opposed to the type header.
+  bool EnableStackFrameInlining = true;
+
+  /// Because we are emitting older C, we cannot specify underlying enum type,
```
pfez: What is older C? State the standard version we're emitting, and the version that would allow to specify the underlying type.

include/revng-c/TypeNames/PTMLCTypeBuilder.h
```diff
@@ -0,0 +1,245 @@
...
+  /// which means this that when we need to preserve enum size across
+  /// the compilation boundary (for example, when editing a type), we need
+  /// to resolve to a trick: we print the maximum value possible - and then use
+  /// it to figure out the original size.
+  /// Setting this flag to true enables printing of such a value.
+  bool EnablePrintingOfTheMaximumEnumValue = false;
+
+  /// All of our structs are packed by definition. Which means every single byte
+  /// has to occupied. In explicit padding mode, the padding is printed as
```
pfez: Typo: "has to occupied". You lost a "be".

include/revng-c/TypeNames/PTMLCTypeBuilder.h
```diff
@@ -0,0 +1,245 @@
...
+    }
+
+    if (StackFrameTypeCache.contains(Key)) {
+      // This is a stack frame.
+      return EnableStackFrameInlining;
+
+    } else {
+      return EnableTypeInlining;
+    }
+  }
```
pfez, line 100: Missing blank line after `}`

include/revng-c/TypeNames/PTMLCTypeBuilder.h
```diff
@@ -0,0 +1,245 @@
...
+  OutStream &out() { return *Out; }
+  void updateTheOutputStream(llvm::raw_ostream &OutputStream) {
+    Out = std::make_unique<OutStream>(OutputStream,
+                                      DecompiledCCodeIndentation,
+                                      isGenerateTagLessPTML());
+  }
+
+public:
+  /// Return a string containing the C type name of \a Type and
+  /// a (possibly empty) \a InstanceName.
```
pfez: What about the other 2 arguments.

include/revng-c/TypeNames/PTMLCTypeBuilder.h
```diff
@@ -0,0 +1,245 @@
...
+  ///       a declaration
+  static bool isDeclarationTheSameAsDefinition(const model::TypeDefinition &T) {
+    return not llvm::isa<model::StructDefinition>(&T)
+           and not llvm::isa<model::UnionDefinition>(&T)
+           and not llvm::isa<model::EnumDefinition>(&T);
+  }
+
+  /// Generates the definition of a new struct type that wraps all the return
+  /// values of \a F.
+  OutStream &generateReturnValueWrapper(const model::RawFunctionDefinition &F);
```
pfez: Missing blank line below.

include/revng-c/HeadersGeneration/PTMLCHeaderBuilder.h
```diff
@@ -0,0 +1,47 @@
...
+
+//
+// This file is distributed under the MIT License. See LICENSE.md for details.
+//
+
+#include "revng-c/TypeNames/PTMLCTypeBuilder.h"
+
+namespace ptml {
+
+class PTMLCHeaderBuilder : public PTMLCTypeBuilder {
```
pfez: I haven't looked into `PTMLCTypeBuilder` yet, but I think this inheritance is weird.

`PTMLCTypeBuilder` seems something that builds types, `PTMLCHeaderBuilder` seems something that builds headers. Headers are not types. In fact, in our generated C headers we do not even emit only types, but also declarations of global variables and functions.

So this definitely needs some rework.
It would probably be better to have `PTMLCHeaderBuilder` hold an instance of a `PTMLCTypeBuilder`.
Can you do that?

pfez: Also, I think in general `PTMLBuilder` (in `revng`) should be called `PTMLTagBuilder` (since it only builds `Tag`s), while these builders you're designing in this patchset should be called something like `PTML*StreamBuilder`, since they have a context, an associated stream, and they are meant to build an entire string that can be parsed as PTML.

include/revng-c/TypeNames/PTMLCTypeBuilder.h
```diff
@@ -0,0 +1,245 @@
...
+
+  /// All of our structs are packed by definition. Which means every single byte
+  /// has to occupied. In explicit padding mode, the padding is printed as
+  /// arrays of bytes. In implicit mode, a `_START_AT` annotation is printed
+  /// instead.
+  bool EnableExplicitPaddingMode = true;
+
+  /// When this flag is set to true, a special `_SIZE($bytes)` annotation is
+  /// printed with every struct definition.
+  bool EnableStructSizeAnnotation = false;
```
pfez: Put these in a separate `struct`.

These are all configurations for printing C code. You could call it `struct CPrintingOptions`.

Also I don't really want these things to be `public`. The constructor should take one of such `struct`s as input, and hold an immutable copy of it.

Or do we really have strong reasons for wanting to be able to mutate these configuration options at runtime? Because in general it seems a bad idea.

include/revng-c/TypeNames/PTMLCTypeBuilder.h
```diff
@@ -0,0 +1,245 @@
...
+  static constexpr std::string_view artificialArrayWrapperPrefix() {
+    return "_artificial_wrapper_";
+  }
+
+  static constexpr std::string_view artificialReturnValueFieldPrefix() {
+    return "field_";
+  }
+  static constexpr std::string_view artificialArrayWrapperFieldName() {
+    return "the_array";
+  }
```
pfez: These probably belong to `PTMLCBuilder`.

include/revng-c/TypeNames/PTMLCTypeBuilder.h
```diff
@@ -0,0 +1,245 @@
...
+    PTMLCBuilder(B),
+    Out(std::make_unique<OutStream>(OutputStream,
+                                    DecompiledCCodeIndentation,
+                                    B.isGenerateTagLessPTML())) {}
+
+  PTMLCTypeBuilder(llvm::raw_ostream &OutputStream, bool GenerateTaglessPTML) :
+    PTMLCBuilder(/* GenerateTagLessPTML = */ GenerateTaglessPTML),
+    Out(std::make_unique<OutStream>(OutputStream,
+                                    DecompiledCCodeIndentation,
+                                    GenerateTaglessPTML)) {}
```
pfez: These constructors do different kind of gymnastic with `PTMLCBuilder` and `GenerateTagLessPTML`  to get the same configuration option for printing C. There should be a single constructor that takes a `const struct CPrintingOptions &Options`, and only uses the `IsTagLessPTML` (or how you wanna call it) option.

include/revng-c/TypeNames/PTMLCTypeBuilder.h
```diff
@@ -0,0 +1,245 @@
...
+
+  OutStream &out() { return *Out; }
+  void updateTheOutputStream(llvm::raw_ostream &OutputStream) {
+    Out = std::make_unique<OutStream>(OutputStream,
+                                      DecompiledCCodeIndentation,
+                                      isGenerateTagLessPTML());
+  }
```
pfez: This is odd. In a perfect world this shouldn't exist.
Why do we even need this in the first place?

The only use is in `DecompileFunction.cpp`. And even there it seems like a workaround.
It seems that what you actually need is a way to precompute the result of `collectInlinableTypes` and then extend its lifetime for use across many functions. And you achieve that but making `collectInlinableTypes` a method of the builder, and letting the builder live longer.

But this seems a hint to the fact that types to inline are actually yet another parameter of the builder itself.

So I propose the following:
1. Let `collectInlinableTypes` not be a method, and let it return its result instead of `void`.
2. Either change the constructor of `PTMLCTypeBuilder` and `PTMLCHeaderBuilder` to take a `const &` to that result, or stuff a `const &` to the result inside the `struct CPrintingOptions` I mentioned above.
3. At this point when you construct a builder you are also configuring it, and doing that is cheap (no copies)
4. This makes the whole dance in `DecompileFunction.cpp` around the builder useless: you can simply `collectInlineableTypes` at the beginning, pass `const &` to its result around as configurations for all the builders you want, just construct a new builder for each function (each builder has shorter lifetime than the result of `collectInlineableTypes`, and they all refer to the same copy, that is never recomputed).
5. `updateTheOutputStream` isn't necessary anymore
6. We can drop the `unique_ptr` around `OutStream`
7. Bonus point. Do we even need the to take the `raw_ostream` as arument in the various constructors? Couldn't we actually switch all the builders to be string-based? I think in principle we could, but in practice we'll not do it, since it would hurt performance (e.g. in some places we serialize directly to file, without first materializing the string and then copying it to a file).

Makes sense?

include/revng-c/TypeNames/PTMLCTypeBuilder.h
pfez: Same holds for the constructors of `PTMLCHeaderBuilder`.

include/revng-c/TypeNames/PTMLCTypeBuilder.h
```diff
@@ -0,0 +1,245 @@
...
+  bool EnableExplicitPaddingMode = true;
+
+  /// When this flag is set to true, a special `_SIZE($bytes)` annotation is
+  /// printed with every struct definition.
+  bool EnableStructSizeAnnotation = false;
+
+  /// Sometimes you don't want to print everything. This lets you specify
+  /// a set of types that will be ignored by \ref typeDefinitions.
+  std::set<model::TypeDefinition::Key> TypesToOmit = {};
+
```
pfez: After thinking about the configuration of the builder. This should probably also be part of it.

include/revng-c/TypeNames/PTMLCTypeBuilder.h
```diff
@@ -0,0 +1,245 @@
...
+    return "field_";
+  }
+  static constexpr std::string_view artificialArrayWrapperFieldName() {
+    return "the_array";
+  }
+
+private:
+  std::optional<DependencyGraph> DependencyCache = std::nullopt;
+  std::set<model::TypeDefinition::Key> TypesToInlineCache = {};
+  std::set<model::TypeDefinition::Key> StackFrameTypeCache = {};
```
pfez: And these two also are part of the configuration.

The configuration I have in mind is rapidly growing in complexity though.
And besides simple `bool` options I see that there's now a bunch of data structure that go together and all have to do with types inlining in 2 flavors:
1. type inlining, i.e. the fact that `struct` and `union` types can have the types of their `struct`, `union` and `enum` fields defined inline, transitively in certain conditions
2. stack type definition inlining, i.e. the fact that the stack types (always struct) have, when possible, their full definition directly in the body of the function

These 2 things are supposed to actually be orthogonal (even if they are not in the old codebase), so they deserve 2 separate `bool` options. And these 2 options should drive the computation of all the ancillary data structure ahead of time (before starting to build C strings), so that the builders can be configured properly with all the information they need before starting.

include/revng-c/TypeNames/PTMLCTypeBuilder.h
```diff
@@ -0,0 +1,245 @@
...
+
+public:
+  /// Gather (and store internally) the list of types that can (and should)
+  /// be inlined. This list is then later used by the invocations of
+  /// \ref printTypeDefinition.
+  void collectInlinableTypes(const model::Binary &Model);
+
+  bool shouldInline(model::TypeDefinition::Key Key) {
```
pfez: This method should be `const`.

include/revng-c/TypeNames/PTMLCTypeBuilder.h
```diff
@@ -0,0 +1,245 @@
...
+  ///       called.
+  OutStream &printFunctionPrototype(const model::TypeDefinition &FunctionType,
+                                    const model::Function &Function,
+                                    bool SingleLine = false);
+  OutStream &printFunctionPrototype(const model::TypeDefinition &FunctionType,
+                                    const model::DynamicFunction &Function,
+                                    bool SingleLine = false);
+  OutStream &printFunctionPrototype(const model::TypeDefinition &FunctionType);
+
+  OutStream &printSegmentType(const model::Segment &Segment);
```
pfez: I find weird that some methods return strings and don't affect the stream while other return a reference to `OutStream`.

I think that overall this is a symptom of the fact that these builders you've designed are not sure if they should manage the stream (which could be a file stream) themselves, or if they just want to build strings and then the serialization of everything to the stream is someone else's job.

If it's the builder who manages the stream (which seems the case, given `unique_ptr`), there shouldn't be any need to give around mutable references to the stream itself.

Did you think about this? What's the actual intent?

pfez (changes requested):
Hey @Cvelth here's the first round of review on the PTML C refactoring.

I think the overall design needs a little more thought, so my comments for now are mainly around that, given that the rest could change sensibly depending on these comments.

I'll take another round once we've settled the main points.

include/revng-c/TypeNames/PTMLCTypeBuilder.h
Cvelth: Ah, yeah, this has nothing to do with ownership or whatnot, it's just to let the users to directly append something _after_ whatever a printer does its job, think
```c++
  B.printFunctionPrototype(/* whatever */) << ";\n";
```

And the fact that the reference is non-const is because it's not a good practice to ever hold const references to a stream, since, in principle, the only useful thing you can do with it is to modify it.

include/revng-c/TypeNames/PTMLCTypeBuilder.h
Cvelth: Done.

include/revng-c/TypeNames/PTMLCTypeBuilder.h
Cvelth: Done.

include/revng-c/TypeNames/PTMLCTypeBuilder.h
Cvelth: Done.

include/revng-c/HeadersGeneration/PTMLCHeaderBuilder.h
Cvelth: 1. This makes some stuff a bit more awkward, but not as big deal overall as I expected. As such, done!

2. I strongly believe that we shouldn't encode `Stream` in the name. Eventually they either all switch to streams or to returning strings (if performance allows). This saves one extra rename when that actually happens.

include/revng-c/TypeNames/PTMLCTypeBuilder.h
Cvelth: My reasoning for doing it this way was for implementation simplicity. But, yeah, I agree that these should be const.

Done.

include/revng-c/TypeNames/PTMLCTypeBuilder.h
Cvelth: Done.

include/revng-c/TypeNames/PTMLCTypeBuilder.h
Cvelth: This is really awkward. But we cannot really encode `GenerateTaglessPTML` in the options. It has to exist at the very highest level - the very basic builder depends on it.

include/revng-c/TypeNames/PTMLCTypeBuilder.h
Cvelth: Done!

include/revng-c/TypeNames/PTMLCTypeBuilder.h
Cvelth: This one, I disagree on. These are cache, not configuration - external world shouldn't really even know about their existence.

For some context, in the new version, the options (as in `bool`s responsible for enabling these) are separate - you can enable one but not the other.

The caches, on the other hand, are not - `TypesToInlineCache` contains _all_ the types. While `StackFrameTypeCache` only shows which of them are stack frames.

There is no deeper reason for this that just that it made implementation easier.

include/revng-c/TypeNames/PTMLCTypeBuilder.h
Cvelth: Done.

include/revng-c/TypeNames/PTMLCTypeBuilder.h
Cvelth: Yeah, this is very awkward, but I still feel like this is the easiest way around it: this lets us fully enclose everything inlining related within the builder, which made it a lot easier to stomach in so many places.

To be honest, I hate the `unique_ptr` no less that you do, but that's a limitation of llvm streams being unmovable blobs of (hopefully) legacy code.

That being said, maybe we can find some kind of middle ground: maybe a builder can be allowed to spawn a sub-builder that just refers to its parent for everything but its stream? Or, even, split it into two: stream aware part and stream agnostic part. But that also feels like quite a workaround.

What do you think?

P. S. Current iteration hasn't touched this, I feel like we should discuss it further first.

include/revng-c/TypeNames/PTMLCTypeBuilder.h
Cvelth: Done.

share/revng/test/tests/model-to-header/inline-enum.h.model.yml.filecheck
```diff
@@ -8,6 +8,5 @@ CHECK: struct _PACKED B {
 CHECK:     enum _ENUM_UNDERLYING(uint32_t) _PACKED A {
 CHECK:         FIRST = 0x0U,
 CHECK:         SECOND = 0x1U,
-CHECK:         _enum_max_value_A = 0xffffffffU,
```
pfez: :thinking: why did this go away?
It's actually necessary for preserving the size of the enum.
Otherwise with `_PACKED` the size of the enum will be 1 byte, while it should be 4 given that `_ENUM_UNDERLYING(uint32_t)`.

share/revng/test/tests/model-to-header/annotate-attibutes.model.yml.filecheck
```diff
@@ -11,7 +11,6 @@
 # CHECK:     prefix_enum_9_none = 0x0U,
 # CHECK:     prefix_enum_9_positive = 0x1U,
 # CHECK:     prefix_enum_9_max_held_value = 0xffffffffU,
-# CHECK:     _enum_max_value_prefix_enum_9 = 0xffffffffffffffffU,
```
pfez: Same here, why did this go away?

lib/Backend/DecompilePipe.cpp
```diff
@@ -29,7 +29,7 @@ void Decompile::run(const pipeline::ExecutionContext &Ctx,
             Module,
             Model,
             DecompiledFunctions,
-            /* GeneratePlainC */ false);
+            /* GeneratePlainC = */ false);
```
pfez: Looks like this escaped the "tagless" rename

lib/Backend/DecompileToDirectoryPipe.cpp
```diff
@@ -52,7 +51,7 @@ void DecompileToDirectory::run(const pipeline::ExecutionContext &Ctx,
               DecompiledFunctions,
               /* GeneratePlainC = */ true);
 
-    ptml::PTMLCBuilder B{ /* GeneratePlainC = */ true };
+    ptml::CBuilder B{ /* GeneratePlainC = */ true };
```
pfez: Actually there's plenty of places that have escaped the "tagless" rename. Please make it consistent everywhere.

include/revng-c/Support/PTMLC.h
```diff
@@ -356,6 +358,25 @@ class CBuilder : public ptml::TagBuilder {
     return keywordTagHelper(toString(TheKeyword));
   }
 
+  ptml::Tag getTypeKeyword(const model::TypeDefinition &T) {
```
pfez: I think this method should be `const`.

include/revng-c/TypeNames/PTMLCTypeBuilder.h
```diff
@@ -0,0 +1,271 @@
...
+    bool EnableExplicitPaddingMode = true;
+
+    /// When this flag is set to true, a special `_SIZE($bytes)` annotation is
+    /// printed with every struct definition.
+    bool EnableStructSizeAnnotation = false;
+
+    /// Sometimes you don't want to print everything. This lets you specify
+    /// a set of types that will be ignored by \ref typeDefinitions.
+    std::set<model::TypeDefinition::Key> TypesToOmit = {};
+  };
```
pfez, line 54: What's the logic you used to pick the default values for the various options?

include/revng-c/TypeNames/PTMLCTypeBuilder.h
```diff
@@ -0,0 +1,271 @@
...
+public:
+  CTypeBuilder(llvm::raw_ostream &OutputStream,
+               CBuilder B,
+               ConfigurationOptions &&Configuration) :
+    CBuilder(B),
+    Out(std::make_unique<OutStream>(OutputStream,
+                                    B,
+                                    DecompiledCCodeIndentation)),
+    Configuration(std::move(Configuration)) {}
+  CTypeBuilder(llvm::raw_ostream &OutputStream, CBuilder B) :
```
pfez: Missing empty line. The indentation is disturbing like this.

include/revng-c/TypeNames/PTMLCTypeBuilder.h
```diff
@@ -0,0 +1,271 @@
...
+    CTypeBuilder(OutputStream, B, ConfigurationOptions{}) {}
+
+  CTypeBuilder(llvm::raw_ostream &OutputStream,
+               bool EnableTaglessMode,
+               ConfigurationOptions &&Configuration) :
+    CTypeBuilder(OutputStream,
+                 ptml::TagBuilder{ .IsInTaglessMode = EnableTaglessMode },
+                 std::move(Configuration)) {}
+  CTypeBuilder(llvm::raw_ostream &OutputStream, bool EnableTaglessMode) :
```
pfez: Missing empty line. The indentation is disturbing like this.

include/revng-c/TypeNames/PTMLCTypeBuilder.h
```diff
@@ -0,0 +1,271 @@
...
+    CTypeBuilder(OutputStream,
+                 ptml::TagBuilder{ .IsInTaglessMode = EnableTaglessMode }) {}
+
+  CTypeBuilder(llvm::raw_ostream &OutputStream,
+               ConfigurationOptions &&Configuration) :
+    CTypeBuilder(OutputStream, {}, std::move(Configuration)) {}
+  CTypeBuilder(llvm::raw_ostream &OutputStream) :
```
pfez: Missing empty line. The indentation is disturbing like this.

include/revng-c/HeadersGeneration/PTMLCHeaderBuilder.h
pfez: 1. :+1: I'm reviewing
2. ok that was just a proposed name to give you the idea. it's fine like it is now

share/revng/test/tests/model-to-header/inline-enum.h.model.yml.filecheck
Cvelth: Wait, this is a good point. Sounds like a bug! Thanks for catching it.

include/revng-c/TypeNames/PTMLCTypeBuilder.h
pfez: I like some of the changes you've done, like separating configurations for type inlining (in other types) and stack frame inlining (in function bodies).

I'll review all of it once again and come back to this.

include/revng-c/TypeNames/PTMLCTypeBuilder.h
Cvelth: As far as I remember, I did my best preserving current values to simplify the transition.

That being said, if you think some of them could use an update, I am not opposed.

include/revng-c/TypeNames/PTMLCTypeBuilder.h
pfez: I see. So it's for letting others append stuff at the end of the stream that is managed by the builder.

This kinda makes sense, but before your refactoring it wasn't really clear if the builder actually owned the stream or not. So I was wary of any situation where passing reference around could further complicate the picture. After the part of the refactoring I've seen, things are much more encapsulated, and the ownership of the stream is clear, so giving mutable references around is not scary anymore.

include/revng-c/TypeNames/PTMLCTypeBuilder.h
pfez: I'll do another full review and we can then resume the conversation.

include/revng-c/TypeNames/PTMLCTypeBuilder.h
pfez: Mmh so perhaps the very basic builder should have a configuration too?

My problem with this thing being passed around is that it's easy to forget about it in one place and end up with some builders that are not homogeneously configured, so you end up generating a mix of tagless and non-tagless stuff.

But the tagless flag is special in this because it's the only thing that is not C specific. And in a way at the moment that's the only configuration option that applies to the base builder. In any case, we can leave it like this for now.

If we ever happen to have a second "configuration option" that should affect also the base builder, we'll wrap it together with the tagless flag. Until then, it can stay like this.

include/revng-c/TypeNames/PTMLCTypeBuilder.h
pfez: I've thought about it and I still think that ideally the builder should be the only one allowed to write stuff in the stream. If the goal is to let users append stuff at the end of the stream, let's just provide an `append` method that takes strings and puts them on the stream.

Let's take a clear stance on this and change the code accordingly. As it is know it's just that the builders aren't sure if they want to manage the stream or not, and it ends up like they only want to do it partially, which is awkward.

pfez: Like, the `out()` method shouldn't even be needed.

Cvelth: Wait, no, `out()` method is a separate issue. That's specifically because of the attempt to reuse the same builder with different streams when emitting functions one by one.

Cvelth: I don't thing it's ever used outside of the builder. And if such a use sneaked in somewhere, `append` or something would easily replace that.

Cvelth: That being said, now that I think about it, append is definitely a good idea! Since we probably want to sanitize such extra input just in case.

include/revng-c/TypeNames/PTMLCTypeBuilder.h
pfez: In any case the caches aren't caches.

My understanding of a cache is something that makes things faster but isn't supposed to alter behavior if something is missing in the cache. The only case where it might alter behavior is when some entry of the cache was supposed to be invalidated but it didn't, so you have a false hit with data that might be stale (hence wrong).

Instead I see:
```c++
  bool shouldInline(model::TypeDefinition::Key Key) const {
    if (not TypesToInlineCache.contains(Key)) {
      // This type is not allowed be inlined.
      return false;
    }
```

If the cache is empty this method says that nothing should be inlined, which is wrong.

So, something of what you say about caches is not really adding up.

Cvelth: In my mind, they are caches in a sense that there's nothing preventing you from computing this list every time you need it (since the rules for selecting such types are pretty straight forward). All they change is that they allow us to only do it once and act like it never changes (which is true for any pipe anyway).

pfez: What I'm trying to say overall is that there is some logic that is external to the builder that has to decide what types (and what stack) can be inlined and what can't. And once that's decided, that info doesn't ever change for the whole lifetime of the builder. So it can effectively be a "configuration" of the builder, or a constructor parameter

Cvelth: That being said, I'm not to attached. If you can think of a better name (or a better system for that matter), I'm all ears!

Cvelth: > What I'm trying to say overall is that there is some logic that is external to the builder

ah, I see the misunderstanding now. The logic is _inside_ the builder now. Hence the cache thing.

It's implicitly built the first time its required.

Cvelth: As for `shouldInline`, I also see the problem now. It was private initially, hence why I didn't bother recomputing the cache there. But then a use case forced me to make it public. My bad on not updating with regards to that.

include/revng-c/TypeNames/PTMLCTypeBuilder.h
pfez: > Wait, no, out() method is a separate issue. That's specifically because of the attempt to reuse the same builder with different streams when emitting functions one by one.

Eh is that actually the case? I think that's the reason why you need `updateTheOutputStream`.
But `out()`? Not convinced.

AFAICT from `git grep "out()"` (which seems to me the only place where it's used) the reason why we need `out()` is to allow clients of the builders to append stuff directly at the end of the stream that is supposed to be managed by the builder.
And I see 2 problems with that:
- that would allow someone to append nasty stuff on the stream without guards (especially if the reference to `Out` escapes somewhere weird) like e.g. non-PTML stuff, unescaped things, potentially anything
- even if that string was valid PTML, we still have a problem if it wasn't generated by the same builder we're calling `out()` on, because now the builder has global persistent state (which is necessay and we can't avoid) that may alter the generation of certain strings

I don't think we can solve both in the scope of this PR, but we can definitely improve the situation.

I think that we should drop `out()` and replace it with 2 different methods:
- `appendRaw`, that takes in a raw string, and depending on the builder state being tagless or not, escape it in HTML
- `appendPTML`, that takes in a string that is expected to be a) a valid PTML string (perhaps we can check?), and b) it should have been created by the same builder instance of the builder you're calling `appendPTML` on (this can't be checked, but at least sticks out in review much more than `out() << whatever`).
It's good that this "need" for `out()` emerges in only one place, but that's not an excuse for settling on something that is still awkward. if condition b) is not true, shit happens, and that should be stated as a "contract" with the user in the comment that describe how to use it.

I'm not in love with the names, but you get the idea.
I'm also not sure `appendRaw` is necessary. If not `append` is enough.

I think `append` instead of `out()` is not perfect but has the following advantages
1. it makes clear already from the name that the builder is in charge of managing the stream
2. it makes it impossible to end up having a reference to the stream escape somewhere where someone can misuse it
3. if you look at current uses of `out()` most of them will become `B.append(B.getstuff())` which makes it a no-brainer to check locally that you're not breaking contract b)
4. for the remaining usages of `out()` we have auxiliary data structures (typically maps) that are populated with `Map[V] = B.getstuff()` and then in another place we do `B.append(Map.at(V))`. so it's not immediate to check for contract b), but as long as all the data injected into the `Map` is generated with `B` it's manually checkable in little time with high accuracy
5. items 3 and 4 seem to cover all uses (AFAICT) of `out()` today and migrate them to `append`. whenever someone will `append` something new, that will stick out in review, and we'll make sure there are no new nasty users, but at least we have a clear contract to check

How does that sound?

Am I missing something?

Cvelth: First of all, I'm all for `out() << X;` -> `append(X);`.

That being said, your `appendRaw` sounds nasty. Mostly because _every_ string is valid PTML string by definition. Tags only carry additional information. But know that I think about it, you probably just meant `sanitizeAndAppend`. Where it's just making sure we don't emit stuff like `<`, `>`and so on. But that can be a bit tricky too, since somebody might try to pass tagged string through that and get confused why it breaks :thinking:

Either way, it feels like we're on the same page for the important bit. We can tweak the small stuff once we see how it ends up looking.

pfez: > you probably just meant `sanitizeAndAppend`. Where it's just making sure we don't emit stuff like `<`, `>` and so on.

Exactly. And to be clear, it should allow passing strings that contain e.g. `<`, but then depending on the tagless mode it should rewrite them as html or not.

> But that can be a bit tricky too, since somebody might try to pass tagged string through that and get confused why it breaks 🤔

Yeah I can see that, and that's why I said that perhaps `appendRaw` (or `sanitizeAndAppend` or even better `escapeAndAppend`) can be left out if it's not really needed in practice.

But if it turns out we can't do without it, then yeah, it's not immune to breakage, but if you end up passing an already escaped string to an API that is called `escapeAndAppend` I don't think you should be surprised if it prints garbage.
If anything, the name is an alert for the reviewer (if not for the developer) that should ask themselves: "are we sure that on this call to `escapeAndAppend` we're not passing in a string that was already escaped?".

I think this is the best we can do without overblowing this PR.

lib/HeadersGeneration/HelpersToHeader.cpp
```diff
@@ -93,71 +92,62 @@ static bool hasUnprintableArgsOrRetTypes(const llvm::Function &F) {
...
+
+    if (Log.isEnabled()) {
+      auto CommentScope = helpers::LineComment(B.out(), B);
+      B.out() << *F.getType();
+    }
+
+    // Print the declaration of the return type, if it's not scalar
+    const auto *RetTy = F.getReturnType();
+    if (auto *RetStructTy = dyn_cast<llvm::StructType>(RetTy)) {
+      printLLVMTypeDeclaration(RetStructTy, F, B);
```
pfez, line 143: In my branch, rebased on current `develop`, this line doesn't compile.

include/revng-c/TypeNames/PTMLCTypeBuilder.h
```diff
@@ -0,0 +1,271 @@
...
+    CTypeBuilder(OutputStream, {}, {}) {}
+
+public:
+  OutStream &out() { return *Out; }
```
pfez: Like we said in another thread. This should go away in favor of an `append` method.

include/revng-c/TypeNames/PTMLCTypeBuilder.h
```diff
@@ -0,0 +1,271 @@
...
+  void updateTheOutputStream(llvm::raw_ostream &OutputStream) {
```
pfez: `setOutputStream` perhaps?

lib/HeadersGeneration/HelpersToHeaderPipe.cpp
```diff
@@ -6,46 +6,70 @@
...
+  llvm::Error checkPrecondition(const pipeline::Context &Ctx) const {
+    return llvm::Error::success();
+  }
+
+  void print(const pipeline::Context &Ctx,
+             llvm::raw_ostream &OS,
+             llvm::ArrayRef<std::string> Names) const {
+    OS << *revng::ResourceFinder.findFile("bin/revng");
+    OS << " helpers-to-header -i=" << Names[0] << " -o=" << Names[1] << "\n";
+  }
```
pfez: I think the `print` method for pipes is supposed to be deprecated. Let me double check how far we are with the deprecation, and what's the plan for removal.

lib/HeadersGeneration/ModelToHeaderPipe.cpp
```diff
@@ -24,42 +23,65 @@ static llvm::cl::opt<bool> InlineTypes("inline-types",
...
+  std::array<pipeline::ContractGroup, 1> getContract() const {
+    using namespace pipeline;
+    using namespace revng::kinds;
+
+    Contract C1(Binary, 0, ModelHeader, 1, InputPreservation::Preserve);
+    return { ContractGroup({ C1 }) };
+  }
+
+  // TODO: BinaryFile here is a placeholder. In principle this pipe has no real
+  // input container. It just juses the model in Ctx to generated HeaderFile.
```
pfez: Typo: `juses`

lib/HeadersGeneration/HelpersToHeaderPipe.cpp
pfez: Ok new directive: blank the body of this method. There's already a branch in the making that will very likely remove the method alltogether.

lib/HeadersGeneration/ModelToHeaderPipe.cpp
```diff
@@ -24,42 +23,65 @@ static llvm::cl::opt<bool> InlineTypes("inline-types",
...
+    return llvm::Error::success();
+  }
+
+  void print(const pipeline::Context &Ctx,
+             llvm::raw_ostream &OS,
+             llvm::ArrayRef<std::string> Names) const {
+    OS << *revng::ResourceFinder.findFile("bin/revng");
+    OS << " model to-header -yaml -i=model.yml -o=" << Names[1] << "\n";
+  }
+};
```
pfez: Make the body of this method empty. It's already informally deprecated and will be removed soon.

lib/HeadersGeneration/HelpersToHeaderPipe.cpp
Cvelth: That being said, I don't think this is a good place to get rid of it. The change here is just white-spaces (I wish github showed that differently in the diffs).

lib/HeadersGeneration/ModelTypeDefinitionPipe.cpp
```diff
@@ -10,37 +10,77 @@
...
+  }
+
+  llvm::Error checkPrecondition(const pipeline::Context &Ctx) const {
+    return llvm::Error::success();
+  }
+
+  void print(const pipeline::Context &Ctx,
+             llvm::raw_ostream &OS,
+             llvm::ArrayRef<std::string> ContainerNames) const {
+    OS << "[CLI tools for pipes are deprecated]\n";
```
pfez: Actually, scratch what I said for the other `print` methods. Make them like this one, at least a user of the deprecated thing has a visual cue.

lib/TypeNames/ModelTypeNames.cpp
```diff
@@ -78,47 +76,46 @@ static std::string getArgumentLocation(llvm::StringRef ArgumentName,
 static std::string
 getArgumentLocationDefinition(llvm::StringRef ArgumentName,
                               const model::DynamicFunction &F,
-                              ptml::PTMLCBuilder &B) {
+                              ptml::CBuilder &B) {
```
pfez: Can't this be `const CBuilder &`?

lib/TypeNames/ModelTypeNames.cpp
```diff
@@ -78,47 +76,46 @@ static std::string getArgumentLocation(llvm::StringRef ArgumentName,
...
   return getArgumentLocation<true>(ArgumentName, F, B);
 }
 
 static std::string getArgumentLocationDefinition(llvm::StringRef ArgumentName,
                                                  const model::Function &F,
-                                                 ptml::PTMLCBuilder &B) {
+                                                 ptml::CBuilder &B) {
```
pfez: Can't this be `const CBuilder &`?

lib/TypeNames/ModelTypeNames.cpp
```diff
@@ -372,27 +368,26 @@ getFunctionAttributesString(const AttributesSet &Attributes) {
...
 
 template<ModelFunction FunctionType>
-static void printFunctionPrototypeImpl(const FunctionType *Function,
+std::string printFunctionPrototypeImpl(const FunctionType *Function,
                                        const model::RawFunctionDefinition &RF,
                                        const llvm::StringRef &FunctionName,
-                                       llvm::raw_ostream &Header,
-                                       ptml::PTMLCBuilder &B,
-                                       const model::Binary &Model,
+                                       ptml::CTypeBuilder &B,
```
pfez: Can't this be `const CBuilder &`? In this case I'm less sure than the previous two. But in general try and look at all the places where you have a `CTypeBuilder &` argument in this file, and double check if you can put a `cont &` instead.

lib/TypeNames/ModelTypeNames.cpp
```diff
@@ -424,31 +419,33 @@ static void printFunctionPrototypeImpl(const FunctionType *Function,
...
 
 template<ModelFunction FunctionType>
-static void printFunctionPrototypeImpl(const FunctionType *Function,
+std::string printFunctionPrototypeImpl(const FunctionType *Function,
                                        const model::CABIFunctionDefinition &CF,
                                        const llvm::StringRef &FunctionName,
-                                       llvm::raw_ostream &Header,
-                                       ptml::PTMLCBuilder &B,
-                                       const model::Binary &Model,
+                                       ptml::CTypeBuilder &B,
```
pfez: Can't this be `const CBuilder &`?

include/revng-c/TypeNames/PTMLCTypeBuilder.h
Cvelth: @pfez ping?

Cvelth: Closing this as per the mattermost confirmation.

include/revng-c/TypeNames/PTMLCTypeBuilder.h
Cvelth: Done.

include/revng-c/TypeNames/PTMLCTypeBuilder.h
Cvelth: Done.

include/revng-c/TypeNames/PTMLCTypeBuilder.h
Cvelth: Done.

lib/HeadersGeneration/HelpersToHeader.cpp
Cvelth: Fixed.

include/revng-c/TypeNames/PTMLCTypeBuilder.h
Cvelth: Done.

lib/HeadersGeneration/ModelToHeaderPipe.cpp
Cvelth: Done.

lib/TypeNames/ModelTypeNames.cpp
Cvelth: Done.

lib/TypeNames/ModelTypeNames.cpp
Cvelth: Done.

lib/TypeNames/ModelTypeNames.cpp
Cvelth: Done.

include/revng-c/TypeNames/PTMLCTypeBuilder.h
Cvelth: Wow, your intuition really did its thing here! Turns out there was indeed a case where `shouldInline` was invoked without first filling the map in!

Cvelth: Done (I added an assert, since it don't think it's viable to make `shouldInline` depend on the model).

include/revng-c/TypeNames/PTMLCTypeBuilder.h
Cvelth: Done.

share/revng/test/tests/model-to-header/inline-enum.h.model.yml.filecheck
Cvelth: Fixed, it was a misconfiguration issue.

lib/Backend/DecompilePipe.cpp
Cvelth: Done.

share/revng/test/tests/model-to-header/annotate-attibutes.model.yml.filecheck
Cvelth: Done.

lib/Backend/DecompileToDirectoryPipe.cpp
Cvelth: Done.

include/revng-c/Support/PTMLC.h
Cvelth: Done.

include/revng-c/TypeNames/PTMLCTypeBuilder.h
Cvelth: Done.

lib/TypeNames/ModelTypeNames.cpp
Cvelth: Done (somehow this escaped closure yesterday).

lib/HeadersGeneration/HelpersToHeaderPipe.cpp
Cvelth: (I didn't touch `pipe::print`s after all since @aleclearmind asked me to touch them as little as possible)

Cvelth: ale's branch went in already, so this is not relevant anymore :tada:

lib/HeadersGeneration/ModelToHeaderPipe.cpp
Cvelth: ale's branch went in already, so this is not relevant anymore :tada:

lib/HeadersGeneration/ModelTypeDefinitionPipe.cpp
Cvelth: ale's branch went in already, so this is not relevant anymore :tada:

PR #33, opened 2024-08-19
BundleEdgesByCycle: implement transformation pass

andrealinux1:
We implement the `BundleEdgesByCycle` LLVMIR transformation pass.

The goal of this transformation is to introduce an header and sink for each node in the graph, in order to enforce the following properties:
- We group the edges incoming into a node, on the basis of the cycle equivalence class. All the edges belonging to the same class, will be grouped as incoming of a new header, which will be in turn connected to the original node.
- We group the edges outgoing from a node, on the basis of the cycle equivalence class. All the edges belonging to the same class, will be grouped as incoming of a new sink node, which will be in turn connected to the original successor.

The goal of this transformation pass is to pre-process the input graph before the `GenericRegionInfo` analysis is run, in order to improve the distinction of different regions.
`GenericRegionInfo` indeed, has a node granularity, and inserting the additional headers and sinks enables us to normalize the graph, by inserting additional nodes for each group of edges that belong to the same cycle equivalence class, and virtually computing the results of the analysis with a _edge_ granularity.

We add some unit tests to check the functionality of this pass.

### DIFF
--- a/include/revng-c/RestructureCFG/BundleEdgesByCycle.h
+++ b/include/revng-c/RestructureCFG/BundleEdgesByCycle.h
@@ -1,0 +1,20 @@
+#pragma once
+
+//
+// Copyright rev.ng Labs Srl. See LICENSE.md for details.
+//
+
+#include "llvm/IR/Function.h"
+#include "llvm/Pass.h"
+
+class BundleEdgesByCyclePass : public llvm::FunctionPass {
+public:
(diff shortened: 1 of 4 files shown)


pfez:
I'm closing this, at least for now, because we've agreed that this specific transformation is not necessary nor beneficial in the new design of the decompilation pipeline.

PR #35, opened 2024-09-13
Centralize creation of local variables

pfez:


pfez:
There aren't any comments here. I'll reopen this on `revng` directly now that they have been merged.

PR #36, opened 2024-09-20
Remove leading `_` from generated names

Cvelth:


lib/HeadersGeneration/ModelToHeader.cpp
```diff
@@ -89,38 +89,38 @@ bool ptml::HeaderBuilder::printModelHeader(const model::Binary &Binary) {
...
     }
   }
 
-  if (not Binary.Segments().empty()) {
+  if (not B.Binary.Segments().empty()) {
     auto Foldable = B.getIndentedScope(CBuilder::Scopes::SegmentDeclarations,
                                        /* Newline = */ true);
     B.appendLineComment("==================");
     B.appendLineComment("==== Segments ====");
     B.appendLineComment("==================");
```
aleclearmind: These are kinda ugly, can you take some time to turn these into something like

```c
/// \defgroup group1 The First Group
/// \{
  ...
/// \}
```

Maybe check if this is valid Doxygen syntax.

lib/ImportFromC/ImportFromCAnalysis.cpp
```diff
@@ -160,11 +159,12 @@ struct ImportFromCAnalysis {
       // declarations.
       if (!ptml::CTypeBuilder::isDeclarationTheSameAsDefinition(*TypeToEdit)) {
         llvm::raw_string_ostream Stream(HeaderConfiguration.PostIncludeSnippet);
-        ptml::CTypeBuilder PI(Stream, /* GenerateTaglessPTML = */ true);
+        ptml::CTypeBuilder PI(Stream, *Model, /* GenerateTaglessPTML = */ true);
         PI.appendLineComment("The type we are editing");
         // The declaration of this type will be near the top of the file.
         PI.printForwardTypeDeclaration(*TypeToEdit);
         PI.append("\n");
+        Stream.flush();
```
aleclearmind: Unnecessary?

lib/PromoteStackPointer/DetectStackSizePass.cpp
```diff
@@ -144,7 +144,11 @@ void DetectStackSize::collectStackBounds(Function &F) {
   // Obtain model::Function corresponding to this llvm::Function
   MetaAddress Entry = getMetaAddressMetadata(&F, "revng.function.entry");
   model::Function &ModelFunction = Binary->Functions().at(Entry);
-  revng_log(Log, "Collecting stack bounds for " << ModelFunction.name().str());
+  if (Log.isEnabled()) {
```
aleclearmind: Unnecessary?
Same elsewhere.

share/revng/test/configuration/revng-c/name-collisions.yml
```diff
@@ -0,0 +1,32 @@
...
+    from:
+      - type: source
+        filter: name-collisions
+    suffix: /
+    command: |-
+      revng model opt -Y -verify "${INPUT}/input-model.yml"
+        1> "${OUTPUT}/model.yml"
+        2> "${OUTPUT}/error.txt"
+
+      && revng artifact emit-model-header /dev/null --model "${OUTPUT}/model.yml"
```
aleclearmind: Use `;` instead of `&&`. We have `set -e`.

share/revng/test/configuration/revng-c/name-collisions.yml
```diff
@@ -0,0 +1,32 @@
...
+        | revng ptml | tee "${OUTPUT}/header.h"
```
aleclearmind: Tests should produce output only if someone is consuming it.
Maybe it's not the best policy, since they always re-run when you run `ninja all`, but that's the current policy.

If no one is consuming the output of this test, don't emit any output.

share/revng/test/configuration/revng-c/name-collisions.yml
```diff
@@ -0,0 +1,32 @@
...
+        | FileCheck "${INPUT}/expected-header.txt"
+
+      || cat "${OUTPUT}/error.txt"
+        | FileCheck "${INPUT}/expected-error.txt"
```
aleclearmind: I'm not sure about the usage of `||`.
If you expect the previous command to fail, ensure that happens with an `if`.

As of now, if the previous command succeeds, the part after `||` won't run (including `FileCheck`) and the test will succeed, if if that's likely not what we want.

share/revng/test/tests/name-collisions/argument-name/input-model.yml
aleclearmind, line 1: Add a comment in each `.yml` stating what you intend to test.

lib/ImportFromC/ImportFromCAnalysis.cpp
Cvelth: Yes, but it was in a different place. This line can be dropped - good catch.

Cvelth: Ah, wait, no, I'm thinking about a different place. This one is necessary. But consider that this object is really short-lived.

lib/PromoteStackPointer/DetectStackSizePass.cpp
Cvelth: Ah, right, the same check is already done inside the macro - my bad.

share/revng/test/configuration/revng-c/name-collisions.yml
Cvelth: I think this came up a lot a lot in the past - this helps debugging quite a bit - since you don't need to partially rerun stuff - outputs are always there. Especially when debugging non-deterministic problems (which caused my to `tee` it).

But I guess if you're adamant about it, it can be dropped now that the bug is fixed.

share/revng/test/configuration/revng-c/name-collisions.yml
Cvelth: > the part after `||` won't run

that's exactly what we want! These tests work like "if command above succeeded, check output against file X, otherwise, check error message against file Y". 

TL;DR: The "doesn't run when successful" part is intentional.

lib/ImportFromC/ImportFromCAnalysis.cpp
aleclearmind: The object is going to go out of scope and get flushed anyway. Why is it necessary?

Cvelth: No idea what, but I saw cases when it only printed a piece of it. Maybe it has something to do with the fact that multiple relevant objects go out of scope there?

share/revng/test/configuration/revng-c/name-collisions.yml
aleclearmind: OK, but it's a bit hard to figure out what the outputs are, given they are non-standard (as opposed to a `--resume` directory, for instance). Maybe add a comment to describe the layout of the output directory.

share/revng/test/configuration/revng-c/name-collisions.yml
aleclearmind: OK, use an `if` and add docs about this being intentional.

share/revng/test/configuration/revng-c/name-collisions.yml
Cvelth: I'm not sure, when you see a directory with `header.h`, `error.txt` and `model.yml`, I feel like it's pretty self-explanatory 🤔

lib/ImportFromC/ImportFromCAnalysis.cpp
aleclearmind: Are you 100% sure? If we need to `.flush()` explicitly I want to know and we need to fix it elsewhere (or in LLVM).
I'd expect the destructor of the stream to do that.

Can you double check?

share/revng/test/configuration/revng-c/name-collisions.yml
aleclearmind: If you run the test it is, if you read its source not so much.

```
# This test produces three files for debugging purposes: ...
```

lib/ImportFromC/ImportFromCAnalysis.cpp
Cvelth: It's not deterministic, so we can never be 100% sure, but I can remove it and run a bunch of times to see if it appears again.

share/revng/test/configuration/revng-c/name-collisions.yml
Cvelth: I guess that makes sense. Will do!

lib/ImportFromC/ImportFromCAnalysis.cpp
aleclearmind: `raw_string_ostream` seems to be unbuffered:
```
  explicit raw_string_ostream(std::string &O) : OS(O) {
    SetUnbuffered();
  }
```

Also, `~raw_ostream` checks that things have been flushed before destruction.
```c++
raw_ostream::~raw_ostream() {
  // raw_ostream's subclasses should take care to flush the buffer
  // in their destructors.
  assert(OutBufCur == OutBufStart &&
         "raw_ostream destructor called with non-empty buffer!");

  if (BufferMode == BufferKind::InternalBuffer)
    delete [] OutBufStart;
}
```

Are you sure they weren't 👻s?

Cvelth: I _really_ want to think there are no 👻s there. But I guess something fishing was happening in there at some point. I'll see if I can reproduce it. Maybe even worth trying to dig for older versions to see what affected it.

Either way, thanks for looking into the stream, this must be builder-related issue then.

lib/ImportFromC/HeaderToModel.cpp
```diff
@@ -24,18 +24,6 @@ static constexpr llvm::StringRef InputCFile = "revng-input.c";
...
-concept HasCustomName = requires(const T &Element) {
-  { Element.CustomName() } -> std::same_as<const model::Identifier &>;
-  { Element.name() } -> std::same_as<model::Identifier>;
-};
-
-template<HasCustomName T>
-static void setCustomName(T &Element, llvm::StringRef NewName) {
-  if (Element.name() != NewName)
-    Element.CustomName() = NewName;
-}
```
pfez, line 37: I'm not comfortable getting rid of this helper. People will forget about the check for sure.

lib/ImportFromC/HeaderToModel.cpp
```diff
@@ -560,7 +550,7 @@ bool DeclVisitor::VisitFunctionDecl(const clang::FunctionDecl *FD) {
       }
 
       model::Argument &NewArgument = FunctionType.Arguments()[Index];
-      setCustomName(NewArgument, FD->getParamDecl(I)->getName());
```
pfez, line 563: Like this line is not 100% equivalent to what was done before. Here the equality check is now gone.

lib/TypeNames/ModelTypeNames.cpp
```diff
@@ -229,38 +229,9 @@ TypeString PCTB::getNamedCInstance(const model::Type &Type,
...
-}
-
 TypeString PCTB::getArrayWrapper(const model::ArrayType &ArrayType) const {
-  std::string Result(artificialArrayWrapperPrefix());
-
-  Result += getArrayWrapperImpl(ArrayType, *this);
-
-  return TypeString(getTag(ptml::tags::Span, std::move(Result)).toString());
+  auto Name = NamingHelper.artificialArrayWrapper(ArrayType);
+  return TypeString(getTag(ptml::tags::Span, std::move(Name)).toString());
```
pfez: Shouldn't the call to `getTag` live in the body of `artificialArrayWrapper`?

Inthe current form, someone reading this needs to look at the body of `artificialArrayWrapper` anyway to understand what's going on.

lib/ImportFromC/HeaderToModel.cpp
pfez: It boils down to the fact that in your helper you no don't have a set of overloaded `name` methods, but you have `type`, `function`, `field` inside.

Any particular reason for that choice?

Because at the moment it is an obstacle to having an helper like this, which is important to prevent frequent programming errors. So what are the advantages of having `type`, `field`, `function` and friends?

pfez: Also, ironically none of the `type`, `field`, `function` method return a type, a field or a function. They all return a name.

Cvelth: Yeah, @aleclearmind already commented on the method names, but thanks for reminding about this helper, it should definitely stay!

lib/ImportFromC/HeaderToModel.cpp
Cvelth: Yeah, my thinking was that in that case it would just override a string with its own copy, not ideal, but not worth the manual check without a helper.

That being said, I'll bring the helper back, so this is not an issue anymore!

lib/TypeNames/ModelTypeNames.cpp
Cvelth: `artificialArrayWrapper` is a part of the model "library". Tag stuff is a part of the PTML "library". I'm not 100% sure which way dependency currently goes, but if we can, I'd really like us to keep them unrelated.

Aside from that, don't forget that `getTag` is also a member of the `TagBuilder` - so putting it inside the naming helper would mean some heavy dependency shifts.

Maybe the core of the issue is the name of `artificialArrayWrapper` member itself? Should I  add `Name` or something? 🤔

pfez: mmh no you are right we don't want to couple the naming with PTML.

yeah add `Name` please. and forget my complains about `getTag`.

lib/HeadersGeneration/ModelToHeader.cpp
Cvelth: Done. I'm not sure I agree this is less ugly, but we can discuss that separately.

lib/ImportFromC/ImportFromCAnalysis.cpp
Cvelth: Done, I guess the other PTML-builder "use-after-free" bug I squashed a while back had some impact on this or something, no idea. Either way, I wasn't able to reproduce this :tada:

lib/PromoteStackPointer/DetectStackSizePass.cpp
Cvelth: Gone!

share/revng/test/configuration/revng-c/name-collisions.yml
Cvelth: Done, please take a look!

share/revng/test/configuration/revng-c/name-collisions.yml
Cvelth: Done, please take a look!

share/revng/test/configuration/revng-c/name-collisions.yml
Cvelth: Done, please take a look!

share/revng/test/tests/name-collisions/argument-name/input-model.yml
Cvelth: I attached them to commands instead, since technically a test could have multiple. That being said, please take a look!

lib/ImportFromC/HeaderToModel.cpp
Cvelth: As discussed on mattermost, I dropped this logic and left a TODO.

lib/ImportFromC/HeaderToModel.cpp
Cvelth: As discussed on mattermost, I dropped this logic and left a TODO.

lib/TypeNames/ModelTypeNames.cpp
Cvelth: Done!

PR #39, opened 2024-10-03
Clift expressions

vasama-work:
New operations:
* `clift.imm`: Integer immediate values.
* `clift.add`, `clift.sub`, `clift.mul`, `clift.div`, `clift.rem`, `clift.neg`: Integer arithmetic operations.
* `clift.not`, `clift.and`, `clift.or`: Logical boolean operations.
* `clift.bitnot`, `clift.bitand`, `clift.bitor`, `clift.bitxor`: Bitwise integer operations.
* `clift.shl`, `clift.shr`: Integer shift operations.
* `clift.eq`, `clift.ne`, `clift.lt` `clift.gt`, `clift.le`, `clift.ge`: Comparison operations.
* `clift.inc`, `clift.dec`, `clift.post_inc`, `clift.post_dec`: Integer increment/decrement operations.
* `clift.cast`: Cast operation.
* `clift.addressof`: Address-of operation.
* `clift.indirection`: Indirection operation (`*x`).
* `clift.assign`: Assignment operation.
* `clift.object_access`: Object access operation (`x.y`).
* `clift.pointer_access`: Indirect object access operation (`x->y`).
* `clift.subscript`: Pointer subscript operation (`x[y]`).
* `clift.comma`: Comma operator (`a, b`).
* `clift.use`: Convert global symbol into an SSA value.
* `clift.call`: Call function or pointer-to-function.

Cast kinds:
* `extend`: Extend integer. Cannot change kind.
* `truncate`: Truncate integer. Cannot change kind.
* `reinterpret`: No-op conversion.
* `decay`: Pointer decay conversion.


include/revng-c/mlir/Dialect/Clift/IR/CliftInterfaces.td
```diff
@@ -78,6 +78,18 @@ def Clift_ValueType : Clift_TypeInterface<"ValueType"> {
              /*methodBody=*/[{}], /*defaultImplemetation=*/[{
               return $_type.getIsConst().getValue();
         }]>,
+        InterfaceMethod<
+            "Returns the unqualified base type of the type.",
```
pfez, line 82: I think this is a copy-paste error from `removeConst`

include/revng-c/mlir/Dialect/Clift/IR/CliftTypes.h
```diff
@@ -83,6 +83,9 @@ class ScalarTupleType : public Type::TypeBase<ScalarTupleType,
   [[nodiscard]] bool getAlias(llvm::raw_ostream &OS) const;
   [[nodiscard]] BoolAttr getIsConst() const;
 
+  [[nodiscard]] clift::ValueType addConst() const { return *this; }
+  [[nodiscard]] clift::ValueType removeConst() const { return *this; }
+
```
pfez, line 197: :thinking: so `ScalarTupleType`s just ignore `const`-qualification

Can you recap the rationale and add it above these 2 methods in a comment please?

lib/mlir/Dialect/Clift/IR/CliftTypes.cpp
```diff
@@ -32,6 +32,20 @@ void CliftDialect::registerTypes() {
            /* End of auto-generated list */>();
 }
 
+static mlir::Type parseConstType(mlir::AsmParser &Parser) {
```
pfez: As stated in the commit message, remember to add printing.

include/revng-c/mlir/Dialect/Clift/IR/CliftOpInterfaces.td
```diff
@@ -15,6 +15,18 @@ def Clift_StatementOpInterface : Clift_OpInterface<"StatementOpInterface"> {
 }
 
 def Clift_ExpressionOpInterface : Clift_OpInterface<"ExpressionOpInterface"> {
+  let methods = [
+    InterfaceMethod<
+      /*desc=*/"Returns the memref operand to read from.",
```
pfez: I think this `desc` is a leftover description from somewhere else.

pfez: Likely `mlir/include/mlir/Dialect/Affine/IR/AffineMemoryOpInterfaces.td:26`

include/revng-c/mlir/Dialect/Clift/IR/CliftOps.td
```diff
@@ -170,18 +174,15 @@ def Clift_MakeLabelOp : Clift_Op<"make_label",
...
   let arguments = (ins Clift_LabelType:$label);
 
   let assemblyFormat = [{
     $label type($label) attr-dict
   }];
 }
 
-def Clift_DoWhileOp : Clift_Op<"do_while",
-                               [Clift_StatementOpInterface,
-                                NoTerminator]> {
```
pfez, line 184: Not really related to this PR, since it was there before, but I think we should thing more thoroughly about the `NoTerminator` trait you've added to many operations.

My intuition at the moment, with the current design of the dialect is that:
- we only have single-block regions both in statements and in expressions
- none of these regions ever require the single block to have a terminator
- `ExpressionRegion`s do `verify` that they end with a `YieldOp`

If these are true, and my reasoning is correct, then we basically have that every clift statement and every clift expression could have the `NoTerminator` trait.
At that point we might very well put it in the `StatementOp` (and later in `ExpressionOp`), don't you think?

include/revng-c/mlir/Dialect/Clift/IR/CliftOps.td
```diff
@@ -74,7 +74,7 @@ def Clift_FunctionOp : Clift_Op<"func",
                                  IsolatedFromAbove,
                                  NoTerminator]> {
   let arguments = (ins StrAttr:$sym_name,
-                       TypeAttrOf<Clift_FunctionType>:$function_type,
+                       TypeAttrOf<Clift_AnyFunctionType>:$function_type,
```
pfez, line 121: Can you remind me the rationale of this rename?

lib/mlir/Dialect/Clift/IR/CliftOps.cpp
```diff
@@ -806,3 +806,227 @@ mlir::LogicalResult YieldOp::verify() {
```
pfez, line 904: Fix the error message please: `must yield an object type or void`.

pfez: Also, restructuring the code as follows would make it clearer to read:
```c++
if (isObjectType(T) or isVoid(T))
  return mlir::success();
  
return emitOpError() << getOperationName() << "must yield void or an object type"'
```

include/revng-c/mlir/Dialect/Clift/IR/CliftTypes.h
vasama-work: They're only used as return types and const return types don't make sense.

include/revng-c/mlir/Dialect/Clift/IR/CliftOps.td
vasama-work: It's just for consistency. I'm naming all type constraints following this pattern. Note that `FunctionType` here is not the type kind attribute `FunctionTypeAttr`. It's a constraint on the concrete type of a *value*.

include/revng-c/mlir/Dialect/Clift/IR/CliftTypes.h
pfez: :+1:

include/revng-c/mlir/Dialect/Clift/IR/CliftEnums.td
```diff
@@ -21,3 +21,12 @@ def Clift_PrimitiveKind: I32EnumAttr<"PrimitiveKind", "primitive type kinds", [
   let genSpecializedAttr = 0;
 }
 
+def Clift_CastKind : I32EnumAttr<"CastKind", "Cast operation semantic kinds", [
+    I32EnumAttrCase<"ZeroExtend",       1,      "zext">,
+    I32EnumAttrCase<"SignExtend",       2,      "sext">,
```
pfez: I know I'm the one who pushed for keeping `zext` and `sext` separate, but I'm having second thoughts.

In most other places (like 2-ary arithmetic operations such as `div` and `rem`) we infer the signedness from the type of the operands. We can pretty much do the same for widening casts and just have `extend` (or `widen`?), which only accepts `Signed` or `Unsigned`, and such that the operand and the return type must have the same `PrimitiveKind`.

It keeps the dialect smaller and leaner, and we still don't have any ambiguity.

lib/mlir/Dialect/Clift/IR/CliftOps.cpp
pfez: Ok this was fixed in a new commit.

lib/mlir/Dialect/Clift/IR/CliftOps.cpp
```diff
@@ -806,3 +806,227 @@ mlir::LogicalResult YieldOp::verify() {
...
+
+  return mlir::success();
+}
+
+//===---------------------------- IndirectionOp ---------------------------===//
+
+mlir::LogicalResult IndirectionOp::verify() {
+  if (isVoid(getResult().getType()))
+    return emitOpError() << getOperationName()
+                         << " cannot dereference a pointer to void.";
```
pfez, line 1017: We also can't dereference pointers to functions

lib/mlir/Dialect/Clift/IR/CliftOps.cpp
```diff
@@ -806,3 +806,227 @@ mlir::LogicalResult YieldOp::verify() {
...
+  auto [UnderlyingType, HasConst] = decomposeTypedef(Type);
+
+  if (HasConst or UnderlyingType.isConst())
+    return emitOpError() << getOperationName()
+                         << " left operand type must be non-const.";
+
+  if (not clift::isLvalueExpression(getLhs()))
+    return emitOpError() << getOperationName()
+                         << " left operand must be an lvalue-expression.";
+
```
pfez, line 1028: :heart_eyes:

pfez: I've been dreaming about this shit on our ad-hoc LLVM monstrosity in the old backend for ages.

lib/mlir/Dialect/Clift/IR/CliftOps.cpp
```diff
@@ -806,3 +806,227 @@ mlir::LogicalResult YieldOp::verify() {
...
+  return mlir::success();
+}
+
+//===--------------------------- ObjectAccessOp ---------------------------===//
+
+static ValueType getField(ValueType ClassType, size_t Index) {
+  const auto GetField = [&](llvm::ArrayRef<FieldAttr> Fields) -> ValueType {
```
pfez, line 1035: Can't you just capture `Index` by copy?

Not very important, but when the set of captures for a lambda is very small I tend to prefer enumerating them explicitly.

lib/mlir/Dialect/Clift/IR/CliftOps.cpp
```diff
@@ -806,3 +806,227 @@ mlir::LogicalResult YieldOp::verify() {
...
+bool ObjectAccessOp::isLvalueExpression() {
+  return clift::isLvalueExpression(getObject());
+}
+
+mlir::LogicalResult ObjectAccessOp::verify() {
+  auto ObjectT = dealias(getObject().getType());
+
+  if (not isClassType(ObjectT))
+    return emitOpError() << getOperationName()
+                         << " operand must have class type.";
```
pfez: I think `struct or union type` is better. As far as clients of the dialect are concerned, there's no such a thing as a class type.

lib/mlir/Dialect/Clift/IR/CliftOps.cpp
```diff
@@ -806,3 +806,227 @@ mlir::LogicalResult YieldOp::verify() {
...
+mlir::LogicalResult PointerAccessOp::verify() {
+  auto PointerT = mlir::dyn_cast<PointerType>(getPointer().getType());
+  if (not PointerT)
+    return emitOpError() << getOperationName()
+                         << " operand must have pointer type.";
+
+  auto PointeeT = PointerT.getPointeeType();
+  if (not isClassType(PointeeT))
+    return emitOpError() << getOperationName()
+                         << " operand type must be a pointer to class type.";
```
pfez: Ditto about "class type".

lib/mlir/Dialect/Clift/IR/CliftOps.cpp
```diff
@@ -806,3 +806,227 @@ mlir::LogicalResult YieldOp::verify() {
...
+mlir::LogicalResult SubscriptOp::verify() {
+  auto PointerT = mlir::dyn_cast<PointerType>(getPointer().getType());
+  if (not PointerT)
+    return emitOpError() << getOperationName()
+                         << " operand must have pointer type.";
+
+  auto PointeeT = PointerT.getPointeeType();
+  if (isVoid(PointeeT))
+    return emitOpError() << getOperationName()
+                         << " cannot dereference a pointer to void.";
```
pfez: We also can't dereference pointers to functions.

include/revng-c/mlir/Dialect/Clift/IR/CliftEnums.td
```diff
@@ -21,3 +21,12 @@ def Clift_PrimitiveKind: I32EnumAttr<"PrimitiveKind", "primitive type kinds", [
...
+    I32EnumAttrCase<"Truncate",         3,      "trunc">,
+    I32EnumAttrCase<"Identity",         4,      "ident">,
```
pfez: I personally don't like the `ident` mnemonic. Other proposals? `bitcast`, `nop`, `reinterpret`?

lib/mlir/Dialect/Clift/IR/CliftOps.cpp
```diff
@@ -806,3 +806,227 @@ mlir::LogicalResult YieldOp::verify() {
...
+      return emitOpError() << " result must have integer type.";
+
+    auto ArgPrimitiveT = mlir::dyn_cast<PrimitiveType>(ArgT);
+    if (not ArgPrimitiveT or not isIntegerKind(ArgPrimitiveT.getKind()))
+      return emitOpError() << " argument must have integer type.";
+
+    if (ResPrimitiveT.getSize() >= ArgPrimitiveT.getSize())
+      return emitOpError() << " result type must be narrower than the argument"
+                              " type.";
+  } break;
```
pfez: In addition to these checks, for `Truncate`, we should check that the `PrimitiveKind` of `ResT` and `ArgT` are the same, otherwise we might have a truncate-cast that also changes signedness.

Also please add a unit test on this.

lib/mlir/Dialect/Clift/IR/CliftTypeHelpers.h
```diff
@@ -64,20 +84,37 @@ inline bool isScalarType(ValueType Type) {
   return mlir::isa<PointerType>(Type);
 }
 
-inline bool isIntegerType(ValueType Type) {
+inline bool isIntegerKind(PrimitiveKind Kind) {
+  switch (Kind) {
+  case PrimitiveKind::NumberKind:
+  case PrimitiveKind::UnsignedKind:
+  case PrimitiveKind::SignedKind:
```
pfez: `PointerOrNumber` should go here too. You can think of it more or less as `uintptr_t`.

pfez: Actually, also `Generic`, because in essence that's also an integer type.

Basically `isIntegerKind` could be implemented just as `return Kind != PrimitiveKind::FloatKind`

lib/mlir/Dialect/Clift/IR/CliftTypeHelpers.h
```diff
@@ -9,17 +9,37 @@
...
+    auto T = mlir::dyn_cast<TypedefTypeAttr>(D.getElementType());
+    if (not T)
+      break;
+
+    Type = T.getUnderlyingType();
+    HasConstTypedef |= D.isConst();
+  }
+
+  return { Type, HasConstTypedef };
+}
```
pfez: So this traverses all the layers of `Typedef`s if there's more than one. Perhaps `traverseTypedefs` is a better name?

pfez: In any case add some Doxygen documentation to stress that it looks through all the layers.

tests/unit/mlir_lit_tests/expressions/arithmetic-add.mlir
```diff
@@ -0,0 +1,32 @@
...
+clift.add %m %m : (!int32_t, !int32_t) -> !int32_t
+
+// CHECK: clift.add [[C]] [[M]] : (!int32_t$const, !int32_t)
+clift.add %c %m : (!int32_t$const, !int32_t)
+// CHECK: clift.add [[C]] [[M]] : (!int32_t$const, !int32_t)
+clift.add %c %m : (!int32_t$const, !int32_t) -> !int32_t
+// CHECK: clift.add [[M]] [[C]] : (!int32_t, !int32_t$const)
+clift.add %m %c : (!int32_t, !int32_t$const)
+// CHECK: clift.add [[M]] [[C]] : (!int32_t, !int32_t$const)
+clift.add %m %c : (!int32_t, !int32_t$const) -> !int32_t
```
pfez: So this thing is just testing that the various deserialization format all work. Nice.

tests/unit/mlir_lit_tests/expressions/cast-identity.mlir
```diff
@@ -0,0 +1,23 @@
+//
+// This file is distributed under the MIT License. See LICENSE.md for details.
+//
+
+// RUN: %revngcliftopt %s
```
pfez: my understanding is that now the `ident` cast works also on enums and typdefs of integer types. can we have tests for those?

lib/mlir/Dialect/Clift/IR/CliftOps.cpp
```diff
@@ -112,6 +112,106 @@ bool clift::impl::verifyExpressionRegion(Region &R, const bool Required) {
...
+    } else if (ValueType Deduced = deduceResultType(Arguments)) {
+      *Result = Deduced;
+    } else {
+      return Parser.emitError(Parser.getCurrentLocation(),
+                              "expected arrow followed by result type");
+    }
+  }
+
+  return mlir::success();
+}
```
pfez, line 179: I think it could be helpful to better document this.

Can you add a comment before this function where you show all the various parsing formats that it allows?

I reconstructed them mentally from the code, and it wasn't a pleasant experience.

Thanks

include/revng-c/mlir/Dialect/Clift/IR/CliftEnums.td
vasama-work: In that case we should apply the same logic to the right shift operation.

lib/mlir/Dialect/Clift/IR/CliftOps.cpp
vasama-work: Why not? https://godbolt.org/z/7P18P3YT9

lib/mlir/Dialect/Clift/IR/CliftOps.cpp
vasama-work: When the scope of the lambda is so limited and its lifetime so well defined, I much prefer the automatic reference capture, because it really doesn't matter. Explicitly listing the captures is just noise that tricks the reader into thinking it actually matters.

lib/mlir/Dialect/Clift/IR/CliftOps.cpp
vasama-work: It's not that we can't dereference functions, but that we cannot perform arithmetic on pointers to functions. I'll change this to check for object type.

include/revng-c/mlir/Dialect/Clift/IR/CliftOps.td
vasama-work: Actually I think we can put `SingleBlock`, `NoRegionArguments` and `NoTerminator` all in `Clift_Op`. These should apply to all of our operations, don't you think?

include/revng-c/mlir/Dialect/Clift/IR/CliftEnums.td
vasama-work: `reinterpret` seems best. If we're merging `zext` and `sext`, should we do:
* `extend`,
* `truncate` (renamed from `trunc`), and
* `reinterpret`?

lib/mlir/Dialect/Clift/IR/CliftTypeHelpers.h
vasama-work: There is `VoidKind` as well.

pfez: Right, I missed that. And that's not an integer :)

include/revng-c/mlir/Dialect/Clift/IR/CliftOps.td
pfez: I agree for `SingleBlock` and `NoTerminator`.

For `NoRegionArguments` I think it should be for sure on `StatementOp` and on `ExpressionOp`. I don't have strong opinions about `GlobalVariableOp`. But I don't think it's appropriate on `FunctionOp`, because my understanding is that function arguments should also be region arguments of the body region. Isn't that the case?

lib/mlir/Dialect/Clift/IR/CliftOps.cpp
pfez: Sure of course you *can*, but after you've done it you can't do anything with the dereferenced value, except of decaying it to a function pointer.

Do we really want that? I don't think so. I don't think there is an assembly construct for which what a user wants to see in C is a dereferenced function pointer.

lib/mlir/Dialect/Clift/IR/CliftOps.cpp
pfez: Right, thanks

include/revng-c/mlir/Dialect/Clift/IR/CliftOps.td
vasama: Yeah, that might be. Turns out `SingleBlock` is not appropriate though, because beyond verifying, it also adds that block to empty regions and we're relying on empty regions in some cases.

vasama-work: Yeah, that might be. Turns out `SingleBlock` is not appropriate though, because beyond verifying, it also adds that block to empty regions and we're relying on empty regions in some cases.

lib/mlir/Dialect/Clift/IR/CliftOps.cpp
vasama-work: I see no harm in allowing it. It doesn't add any complexity. And whenever we make that C to Clift compiler, we're going to encounter C that does this.

pfez: Fair enough. Let's leave this like this.

include/revng-c/mlir/Dialect/Clift/IR/CliftOps.td
pfez: Oh I see. Can we do `SingleBlockOrEmpty` instead?

lib/mlir/Dialect/Clift/IR/CliftOps.cpp
pfez: I always prefer to restrict as much as possible access to external context inside a lambda. And capturing everything by reference doesn't restrict anything.

But I agree that this lambda is so small that it makes little difference.

Like I said, not very important. Do what you prefer in this case, and close this thread when you feel like it.

include/revng-c/mlir/Dialect/Clift/IR/CliftEnums.td
pfez: as long as the signed/unsigned types remain spelled at least once in the textual assembly form, yeah

include/revng-c/mlir/Dialect/Clift/IR/CliftEnums.td
pfez: perfect

include/revng-c/mlir/Dialect/Clift/IR/CliftOps.td
vasama-work: It doesn't seem necessary. All of our operation regions are either `StatementRegion`s or `ExpressionRegion`s, which do the verification already, except for the `Clift_ModuleOp` region, and in that case `SingleBlock` is appropriate and present.

lib/mlir/Dialect/Clift/IR/CliftTypes.cpp
vasama-work: We decided to remove `!clift.const` instead.

include/revng-c/mlir/Dialect/Clift/IR/CliftOps.td
pfez: :+1:

include/revng-c/mlir/Dialect/Clift/IR/CliftOps.td
```diff
@@ -113,15 +113,6 @@ def Clift_FunctionOp : Clift_Op<"func",
...
 
-    mlir::ArrayAttr getArgAttrsAttr();
-    mlir::ArrayAttr getResAttrsAttr();
-
-    void setArgAttrsAttr(mlir::ArrayAttr attrs);
-    void setResAttrsAttr(mlir::ArrayAttr attrs);
-
-    mlir::Attribute removeArgAttrsAttr();
-    mlir::Attribute removeResAttrsAttr();
-
```
pfez, line 124: Hey where did this methods go? I thought they were necessary for `CallableOpInterface` or `FunctionOpInterface`. What happened?

tests/unit/mlir_lit_tests/operations/function-with-mistyped-return.mlir
```diff
@@ -15,7 +15,7 @@
 
 // CHECK: cannot return expression in function returning void
 clift.module {
-  clift.func "f" !f {
```
pfez, line 18: Nothing this here for lack of a better place.
Can you add tests where we check custom parsing and printing of function?
It would be nice to test some of the corner cases with various kinds of arguments, and possibly attributes too. IIUC at the moment in the tests we only have functions that take no arguments and return void, which is unfortunate.

lib/mlir/Dialect/Clift/IR/CliftOps.cpp
```diff
@@ -1110,13 +1119,188 @@ mlir::LogicalResult SubscriptOp::verify() {
...
 }
+
+//===-------------------------------- UseOp -------------------------------===//
+
+mlir::LogicalResult
+UseOp::verifySymbolUses(SymbolTableCollection &SymbolTable) {
+  auto Module = getOperation()->getParentOfType<clift::ModuleOp>();
+  Operation *Op = SymbolTable.lookupSymbolIn(Module, getSymbolNameAttr());
+
+  /*  */ if (auto V = mlir::dyn_cast_or_null<GlobalVariableOp>(Op)) {
```
pfez: Drop the leftover leading comment

lib/mlir/Dialect/Clift/IR/CliftOps.cpp
```diff
@@ -1110,13 +1119,188 @@ mlir::LogicalResult SubscriptOp::verify() {
...
+
+  if (Parser.parseOptionalKeyword("as").succeeded()) {
+    if (Parser.parseLParen().failed())
+      return mlir::failure();
+
+    if (Parser.parseTypeList(ArgumentTypes).failed())
+      return mlir::failure();
+
+    if (Parser.parseRParen().failed())
+      return mlir::failure();
```
pfez: There is no test for this part of the syntax. Please add some.

lib/mlir/Dialect/Clift/IR/CliftOps.cpp
```diff
@@ -1110,13 +1119,188 @@ mlir::LogicalResult SubscriptOp::verify() {
...
+                              " of the function, ignoring qualifiers.";
+  }
+
+  auto ReturnT = mlir::cast<clift::ValueType>(FunctionTypeAttr.getReturnType());
+  auto ResultT = mlir::cast<clift::ValueType>(getResult().getType());
+
+  if (ResultT != ReturnT.removeConst())
+    return emitOpError() << getOperationName()
+                         << " result type must match the return type of the"
+                            " function, ignoring qualifiers.";
```
pfez: I'd also like to have a test for each of those expected failures.

pfez (changes requested):

include/revng-c/mlir/Dialect/Clift/IR/CliftOps.td
vasama-work: I added the formal attributes:
```
                       OptionalAttr<DictArrayAttr>:$arg_attrs,
                       OptionalAttr<DictArrayAttr>:$res_attrs,
```

The formal attributes cause these functions to be defined. Previously we were just pretending to have such attributes.

tests/unit/mlir_lit_tests/operations/function-with-mistyped-return.mlir
vasama-work: Functions also get tested by the various expression tests, especially the `CallOp` tests. I added another `FunctionOp` test containing argument and result attributes.

pfez: Ok thanks.

include/revng-c/mlir/Dialect/Clift/IR/CliftOps.td
pfez: Neat.

Any particular reason why we didn't think about adding the formal attributes sooner? Or is this just the proper way to do it and you only found out now?

vasama-work: I think we probably could have done this originally, but we didn't have syntax for specifying the attributes until now.

PR #43, opened 2024-10-30
ScopeGraph: introduce ScopeGraph

andrealinux1:
Introduce the `ScopeGraph` view over an `llvm::Function`, and the necessary utilities to have `Dashed` and `Goto` edges over it.


include/revng-c/RestructureCFG/GeneratorIterator.h
```diff
@@ -0,0 +1,137 @@
...
+#include "revng/Support/Generator.h"
+#include "revng/Support/IRHelpers.h"
+
+template<typename T>
+class GeneratorIterator {
+public:
+  // We need the specification of the following `iterator_traits` so that
+  // `llvm::GraphWriter` doesn't complain in trying to perform `std::distance`
+  // over the `child_begin`/`child_end` traits
+  using difference_type = unsigned;
```
pfez: Why `unsigned`?

It should be a signed integer for sure.
Imagine you have 2 iterators `i` and `j = std::next(i)`.
You want to be able to compute:
- `j - i` (1)
- `i - j` (-1)

I'd go for `ptrdiff_t`.

include/revng-c/RestructureCFG/GeneratorIterator.h
```diff
@@ -0,0 +1,137 @@
...
+public:
+  bool IsEnd = false;
+  mutable bool IsDead = false;
+  mutable cppcoro::generator<T> Coroutine;
+  inner_iterator Begin;
+  inner_iterator End;
+
+  // We assume that `T` is copy constructible and cheap to copy. These
+  // assumptions are inherited by the assumptions made on `NodeRef`, since the
+  // only use we envision is on those objects.
```
pfez: Turn these assumptions into `concept`s, and use them to constrain `T`.

`std::is_trivially_copyable` a `std::is_copy_constructible` and `std::is_copy_assignable` might help.

pfez: If you manage to do this, you can move this header out of here, and in `include/ADT`

include/revng-c/RestructureCFG/GeneratorIterator.h
```diff
@@ -0,0 +1,137 @@
...
+  GeneratorIterator(cppcoro::generator<T> &&Coroutine) :
+    IsEnd(false),
+    IsDead(false),
+    Coroutine(std::move(Coroutine)),
+    Begin(this->Coroutine.begin()),
+    End(this->Coroutine.end()) {
+    updateContent();
+  }
+
+  GeneratorIterator(const GeneratorIterator &Other) { *this = Other; }
```
pfez: Why the copy constructor and not the move constructor?

You have the move assignment below, but not the move assignment. Weird...

include/revng-c/RestructureCFG/GeneratorIterator.h
```diff
@@ -0,0 +1,137 @@
...
+    // `operator`, in order to create a copy of the `GeneratorIterator` object,
+    // which we leave in the `IsDead` state, this object will be just
+    // dereferenced (returning the value that we pre-computed into `Content`).
+    // Therefore it should not be incremented.
+    Begin = Other.Begin;
+    End = Other.End;
+    Content = Other.Content;
+
+    return *this;
+  }
```
pfez: The comment is not clear and it doesn't seem to match what happens.
It says that the `GeneratorIterator` is left in `IsDead` state, but
- I think it should say `Other` instead of `GeneratorIterator`
- it doesn't set it to `IsDead` state

include/revng-c/RestructureCFG/GeneratorIterator.h
```diff
@@ -0,0 +1,137 @@
...
+    IsDead = Other.IsDead;
+    Coroutine = std::move(Other.Coroutine);
+    Begin = std::move(Other.Begin);
+    End = std::move(Other.End);
+    Content = std::move(Other.Content);
+
+    // After invoking the move assignment, we mark the `Other` object as dead
+    Other.IsDead = true;
+
+    return *this;
```
pfez:
```c++
if (this != &Other) {
  // current body here
}
return *this
```

Otherwise if `this == &Other` you're hopelessly setting it do `IsDead` state.

include/revng-c/RestructureCFG/GeneratorIterator.h
```diff
@@ -0,0 +1,137 @@
...
+    // It is correct that we do not perform a check on the `IsDead` state before
+    // returning `Content`, since it is expected that a `GeneratorIterator`
+    // object left in a `IsDead` state by, e.g., the copy constructor, can be
+    // dereferenced (returning the value in `Content`).
+    return Content;
+  }
+
+private:
+  // Private method to update the `Content` field by dereferencing `Begin`. If
+  // we are at the end of the `internal_iterator`, we also set the `IsDead`
```
pfez: I don't see the `IsDead` flag being set.

Is the comment old, or you forgot something?

include/revng-c/RestructureCFG/GeneratorIterator.h
```diff
@@ -0,0 +1,137 @@
...
+  using pointer = T *;
+  using reference = T &;
+  using iterator_category = typename std::forward_iterator_tag;
+
+public:
+  using inner_iterator = cppcoro::generator<T>::iterator;
+
+public:
+  bool IsEnd = false;
+  mutable bool IsDead = false;
```
pfez, line 32: Removing `mutable` from this effectively compiles. I think it's just because you forgot to set `IsDead` in at least one place where `mutable` would have been necessary. But please at the end of the review double check if this is really necessary.

include/revng-c/RestructureCFG/GeneratorIterator.h
```diff
@@ -0,0 +1,137 @@
...
+  mutable cppcoro::generator<T> Coroutine;
```
pfez, line 33: Also removing `mutable` from here still compiles.

I'm puzzled, why is `mutable` even needed?

include/revng-c/RestructureCFG/GeneratorIterator.h
pfez: Overall I have some trouble understanding if `GeneratorIterator` is supposed to be:

1. copyable? the fact that it has a `cppcoro::generator` as data member seems to suggest that it souldn't be possible to copy it, because `cppcoro::generator` isn't copyable. but then you go and implement copy constructor and copy assignment. my current understanding is that in principle it shouldn't be copyable, but then there's some code we don't control in LLVM that *needs* to make a copy. Can you explain this in detail?

2. movable? I think that it should be movable. and for extra safety it would be nice that moving it caused the `Coroutine` to be moved too (otherwise you could resume it twice) and set the moved `IsDead` to `true`. Why is the last thing necessary is not clear. Can you explain?

pfez: But I like the fact that the implementation of the copy constructor uses the copy assignment. When you implement the move constructor, please make sure that the implementation uses the move assignment.

include/revng-c/RestructureCFG/GeneratorIterator.h
```diff
@@ -0,0 +1,137 @@
...
+    // dereferenced (returning the value in `Content`).
+    return Content;
+  }
+
+private:
+  // Private method to update the `Content` field by dereferencing `Begin`. If
+  // we are at the end of the `internal_iterator`, we also set the `IsDead`
+  // flag.
+  void updateContent() {
+    if (Begin != End) {
```
pfez: For extra safety, assert that we're not in `IsDead` state here.

include/revng-c/RestructureCFG/GeneratorIterator.h
```diff
@@ -0,0 +1,137 @@
...
+
+public:
+  bool operator==(const GeneratorIterator &Other) const {
+    if (IsEnd)
+      return Other.Begin == Other.End;
+    if (Other.IsEnd)
+      return Begin == End;
+    else
+      return Begin == Other.Begin;
+  }
```
pfez: This comparison operator doesn't handle the `IsDead` case.

Also, I have some doubts about the `IsEnd` state.

Finally, in the final `else` you don't check the `End`s.

So can you explain in more detail how is this supposed to work?

I want to make sure all the corner cases are handled, and that their handling make sense.

include/revng-c/RestructureCFG/GeneratorIterator.h
```diff
@@ -0,0 +1,137 @@
...
+  // Private method to update the `Content` field by dereferencing `Begin`. If
+  // we are at the end of the `internal_iterator`, we also set the `IsDead`
+  // flag.
+  void updateContent() {
+    if (Begin != End) {
+      Content = *Begin;
+    } else {
+
+      // This is probably superfluous, since the `==` operator will already
+      // handle the `Begin == End` per se
```
pfez: Drop this comment: it nudges you in wanting to drop the next line. Which superfluous now, but it may become necessary depending on how you change `operator==`. So drop the comment and leave the next line.

include/revng-c/RestructureCFG/GotoTargetGraphTraits.h
```diff
@@ -0,0 +1,108 @@
...
+  }
+
+  // We then move to returning the additional successor represented by the
+  // `ScopeCloser` edge, if present at all
+  llvm::BasicBlock *GotoTarget = getGotoTarget(BB);
+  if (GotoTarget) {
+    co_yield GotoTarget;
+  }
+
+  // TODO: co_return end iterator
```
pfez, line 29: This TODO is still not implemented.

include/revng-c/RestructureCFG/ScopeCloserGraphTraits.h
```diff
@@ -0,0 +1,108 @@
...
+  }
+
+  // We then move to returning the additional successor represented by the
+  // `ScopeCloser` edge, if present at all
+  llvm::BasicBlock *ScopeCloserTarget = getScopeCloser(BB);
+  if (ScopeCloserTarget) {
+    co_yield ScopeCloserTarget;
+  }
+
+  // TODO: co_return end iterator
```
pfez, line 29: This TODO is still not implemented.

include/revng-c/RestructureCFG/MarkerCallUtils.h
```diff
@@ -0,0 +1,145 @@
...
+    MarkerCallFunction->setLinkage(llvm::GlobalValue::ExternalLinkage);
+    MarkerCallFunction->addFnAttr(llvm::Attribute::OptimizeNone);
+    MarkerCallFunction->addFnAttr(llvm::Attribute::NoInline);
+    MarkerCallFunction->addFnAttr(llvm::Attribute::NoMerge);
+    MarkerCallFunction->addFnAttr(llvm::Attribute::NoUnwind);
+    MarkerCallFunction->addFnAttr(llvm::Attribute::WillReturn);
+    MarkerCallFunction
+      ->setMemoryEffects(llvm::MemoryEffects::inaccessibleMemOnly()
+                         | llvm::MemoryEffects::readOnly());
+  }
```
pfez: Why are markers still identified by name?

Last time we spoke we said to use `OpaqueFunctionPool`s.

include/revng-c/RestructureCFG/GotoTargetGraphTraits.h
```diff
@@ -0,0 +1,108 @@
...
+template<class GraphType>
+struct Goto {
+  const GraphType &Graph;
+
+  inline Goto(const GraphType &G) : Graph(G) {}
+};
+
+/// Specializes `GraphTraits<Goto<T *>>`
+template<>
+struct llvm::GraphTraits<Goto<llvm::BasicBlock *>> {
```
pfez, line 44: This is really not what REP 20 says.

You've made 2 different `GraphTrait`s:
1. one for `Goto`, that currently sees `goto` edges *in addition* to regular successors 
2. one for `ScopeCloser`, that currently seed scope closer edges *in addition* to regular successors


REP 20 only talks about a *single* view (the `ScopeGraph) on a CFG that has to do *both* the following
1. *ignore* outgoing edges for blocks that have a `Goto` marker.
2. *see* `ScopeCloser` annotation as edges, even if they don't apper on the CFG.

In addition, a `Goto` marker is only allowed on a block with a single outgoing edge, otherwise the `Goto` marker should also need a second argument to tell which successor is marked as goto and should be ignored by the view.
This condition is not checked in the `verify` and in fact there's even a unit test that breaks it, without failing.

The single view described in REP 20, the `ScopeGraph`, should be the only `GraphTrait` we implement. AFAICT we never need a view that only does something on `Goto` and not on `ScopeCloser` or viceversa.

include/revng-c/RestructureCFG/GotoTargetGraphTraits.h
```diff
@@ -0,0 +1,108 @@
...
+  : public llvm::DefaultDOTGraphTraits {
+  using llvm::DefaultDOTGraphTraits::DefaultDOTGraphTraits;
+
+  std::string getNodeLabel(const llvm::BasicBlock *N,
+                           const Goto<llvm::Function *> G) {
+    return N->getName().str();
+  }
+
+  // TODO: we may want to specialize the `getEdgeAttributes` method in order to
+  //       print the `GotoTarget` edges as dashed
```
pfez, line 107: This is a cool idea but:
1. first of all you really have to make a single `GraphTrait` representing the `ScopeGraph`
2. then I want the following customizations for:
  - the CFG edges that are "ignored" due to the fact that they are cancelled by a `Goto` marker, should be dashed, and should be marked in graphviz as `[constraint=false]` [as explained here](https://stackoverflow.com/questions/22756929/graphviz-make-edges-not-affecting-the-hierarchy-of-nodes) so that they don't affect the layout of the graph
  - the `ScopeCloser` edges should be marked in a different color

include/revng-c/RestructureCFG/MarkerCallUtils.h
pfez: I'm not reviewing the rest of the `MarkerCallBuilder` until we clarify this point.

pfez (changes requested):

include/revng-c/RestructureCFG/GeneratorIterator.h
andrealinux1: Right, fixed here https://github.com/revng/revng/blob/d8ada1163cf18b087655d87782b79c1a120885c9/include/revng/ADT/GeneratorIterator.h#L27

include/revng-c/RestructureCFG/GeneratorIterator.h
andrealinux1: Defined a new concept with your suggestions here https://github.com/revng/revng/blob/d8ada1163cf18b087655d87782b79c1a120885c9/include/revng/ADT/GeneratorIterator.h#L16, and moved this header to `include/ADT`

include/revng-c/RestructureCFG/GeneratorIterator.h
andrealinux1: The new copy assignment is implemented here https://github.com/revng/revng/blob/d8ada1163cf18b087655d87782b79c1a120885c9/include/revng/ADT/GeneratorIterator.h#L82, but I guess you'll just go through it again in the new review

include/revng-c/RestructureCFG/GeneratorIterator.h
andrealinux1: The move assignment has gone, we use the default implementation, added a comment on this here https://github.com/revng/revng/blob/d8ada1163cf18b087655d87782b79c1a120885c9/include/revng/ADT/GeneratorIterator.h#L91

include/revng-c/RestructureCFG/GeneratorIterator.h
andrealinux1: This method and the pre-dereferentiation mechanism has been removed now

include/revng-c/RestructureCFG/GeneratorIterator.h
andrealinux1: `mutable` is necessary when invoking the copy constructor, because we do something like `Other.IsDead = true;`. The code correctly compiles without it because in the current code there is no use of the `Scope` view which invokes the copy constructor.

include/revng-c/RestructureCFG/GeneratorIterator.h
andrealinux1: `mutable` is necessary when invoking the copy constructor, because we do something like `Other.IsDead = true;`. The code correctly compiles without it because in the current code there is no use of the `Scope` view which invokes the copy constructor.

include/revng-c/RestructureCFG/GeneratorIterator.h
andrealinux1: This method doesn't exists anymore, but now every method checks the `IsDead` state as first thing

include/revng-c/RestructureCFG/GeneratorIterator.h
andrealinux1: The `IsDead` is now checked at the beginning of this method https://github.com/revng/revng/blob/d8ada1163cf18b087655d87782b79c1a120885c9/include/revng/ADT/GeneratorIterator.h#L97, and I added a comment here https://github.com/revng/revng/blob/d8ada1163cf18b087655d87782b79c1a120885c9/include/revng/ADT/GeneratorIterator.h#L104 explaining why we special case the situation where either `this` or `Other` is a sentinel end iterator

include/revng-c/RestructureCFG/GeneratorIterator.h
andrealinux1: This has been dropped

include/revng-c/RestructureCFG/ScopeCloserGraphTraits.h
andrealinux1: I now `co_return` here https://github.com/revng/revng/blob/d8ada1163cf18b087655d87782b79c1a120885c9/include/revng/RestructureCFG/ScopeGraphGraphTraits.h#L35

include/revng-c/RestructureCFG/GotoTargetGraphTraits.h
andrealinux1: This `GraphTrait` disappeared

include/revng-c/RestructureCFG/MarkerCallUtils.h
andrealinux1: We agreed on using the `FunctionTag` mechanism, this is now implemented and specifically we use this routine https://github.com/revng/revng/blob/d8ada1163cf18b087655d87782b79c1a120885c9/include/revng/RestructureCFG/MarkerCallUtils.h#L21 to perform the `MarkerCallFunction` retrieval

include/revng-c/RestructureCFG/GotoTargetGraphTraits.h
andrealinux1: Agree with your observations on how we would implement the graph, but this cannot be done at the moment due to how the `DotGraphWriter` wants to copy the edge iterators, something that we cannot do at the moment. Indeed this TODO was more a "we may want to do this in the future", in order to not loose track of the idea. I added to the comment your precisation on how we would mark the edges with different colors and the `constraint` thing here https://github.com/revng/revng/blob/8fbde7a81df05ca5e3c9ec29a0d416dc599e414f/include/revng/RestructureCFG/ScopeGraphGraphTraits.h#L115, but I would say we may want to not tackle this now.

include/revng-c/RestructureCFG/GeneratorIterator.h
andrealinux1: I think we discussed this a lot in person last time, so I guess most of the doubts here have been already resolved, and it may be more useful to comment the new version on the new PR, but let me know if you want some extra clarifications on this point

include/revng-c/RestructureCFG/GotoTargetGraphTraits.h
andrealinux1: You are right, as we discussed in person on this last time. The new implementation should reflect what we discussed, so I think you'll directly comment on that on the new PR.

pfez:
I'm closing this since this PR has already been migrated to revng

